PR #5, opened 2020-12-11
Create midi.hexpat

mathaou:
This is for single track MIDI files in format 0. The pattern must be modified to detect specific patterns since I don't know C at all.


mathaou:
Included README edit

WerWolv:
Thanks a lot :)

PR #8, opened 2021-01-09
Add UEFI structs

Foxboron:
This is a draft PR that includes the current progress on parsing UEFI structs as defined in [Version 2.8 Errata A](https://uefi.org/sites/default/files/resources/UEFI_Spec_2_8_A_Feb14.pdf). Attempting to parse the revocation list published by UEFI as they are easy to fetch and usually interesting to look at.

https://uefi.org/revocationlistfile

Currently it hits the wall of what the pattern language can do, but nonetheless probably interesting.

Signed-off-by: Morten Linderud <morten@linderud.pw>


WerWolv:
Thanks a lot for the PR!

What are some things that are currently missing in the language in order to properly parse the data? I'm always glad to add new features.

Foxboron:
A `EFI_SIGNATURE_LIST` is one or more signature lists. Currently I can only have one `EFI_SIGNATURE_LIST` struct in `dbx_esl`, and no way to say "read EFI_SIGNATURE_LIST until EOF". That's the first one.

The second one is that each `EFI_SIGNATURE_LIST` can contain multiple signatures inside the list (e.g the published DBX has 3 signature lists. 2 signature lists with one x509 certificate, and one signature list with 190 sha256 checksums). I know the size of the last signature list with the checksums, and I know how large each of the sha256 checksum structs are. But I can't say "Parse this struct (200/20) times".

`sizeof` would be neat, but defining the static sizes isn't really hard either.

Thanks for the work on this project :)

WerWolv:
The first issue requires some kind of loops and a way to get the current offset and the size of the whole file probably. That's definitely something I can implement.
For the second one, aren't if-else conditionals enough for that? 

sizeof and co are planed and will probably come with the next version :)

Foxboron:
>The first issue requires some kind of loops and a way to get the current offset and the size of the whole file probably. 

`EFI_SIGNATURE_LIST x509_1[...];`

You could maybe also do it declaratively by saying the above. Would allow `[5...]` which would be "read at least 5 until EOF"

>For the second one, aren't if-else conditionals enough for that?

No, I need to be able to express "read 10 of these structs" with math in practise.

UserBlackBox:
> The first issue requires some kind of loops and a way to get the current offset and the size of the whole file probably. That's definitely something I can implement.

This would be much appreciated for writing patterns for filetypes like PNG which have an unknown amount of chunks.

Foxboron:
For the sake of inspiration I implemented the same with GNU Poke

https://github.com/Foxboron/poke-uefi/blob/master/uefi.pk#L87

`EFI_SIGNATURE_LIST[] lists` was enough to read the lists until EOF and similar to the suggested syntax above

WerWolv:
I'm just gonna merge this PR now because what the hell

PR #9, opened 2021-01-17
Improved PE patterns for both x86 and x64 files.

ThisALV:
Added sections table and data directories.
Support for 64bits binaries.
Separated files for 32bits and 64bits binaries.

EDIT : I had to make coffHeaderPointer an union so I can use the address to locate the sections table.


WerWolv:
Thanks a lot for the PR!
Having support for both 32 and 64 bit is great. One issue I see though with separating the two is that ImHex will only ask the user if they want to use the 32 bit one since the MIME type is the same. How hard would it be to combine both of them into one? The latest version supports `if` statements within structs so it might be possible if the architecture is encoded into the format.
Never the less I should probably improve the popup in ImHex too.

ThisALV:
"ImHex will only ask the user if they want to use the 32 bit one since the MIME type is the same" which popup ? I just have to open the executable file and the pattern file with the imgui file browser. Is there a feature I missed ?

WerWolv:
Yeah, if you have patterns inside the `patterns` folder next to the executable, every time you open a file it will check if its MIME type matches the one specified in any pattern file in there using `#pragma MIME something`. If it matches, ImHex will ask the user if they want to use that pattern to look at the data. That's the reason why that pragma exists

ThisALV:
You mean that I have to create a `patterns` (or `pattern`) folder next to imhex.exe and put the *.hexpat here ? Or you mean *.hexpat files next to imhex.exe ?
Anyvays I tried each one and none of these methods actually work.

ThisALV:
Even I couldn't make the feature you were talking about work, I merged the two files into one with automatic detection for 64bits (PE32+ format) mode.

WerWolv:
Sorry. This is the behaviour I was talking about:



And I really have to improve it because it only allows the first one found to be loaded

patterns/pe.hexpat
```diff
@@ -104,9 +148,31 @@ struct DOSStub {
 	u8 data[11];
 };
 
+union SectionMisc {
+	u32 physicalAddress;
+	u32 virtualSize;
+};
+
+struct Section {
+	s8 name[8];
```
WerWolv, line 157: Instead of `s8`, `char` can be used here. A `char` array will be displayed as string instead of an array in the pattern data view

patterns/pe.hexpat
```diff
@@ -104,9 +148,31 @@ struct DOSStub {
...
 	DOSStub dosStub;
 };
 
-PEHeader peHeader @ 0x00;
\ No newline at end of file
+PEHeader peHeader @ 0x00;
+
+#define COFF_HEADER peHeader.dosHeader.coffHeaderPointer
+
+Section sectionsTable[COFF_HEADER.numberOfSections] @ nextAfter("COFF_HEADER");
```
WerWolv, line 178: oof this looks so weird but works because my preprocessor isn't super sophisticated. Maybe put the `"` into the define as well. That will make it look more natural

WerWolv (changes requested):
All in all it looks great, thanks a lot :)

patterns/pe.hexpat
ThisALV: Unfortunately I actually can't because I would not be able to access the numberOfSections field that way. I'm just gonna delete the define, the string isn't very long anyway.

WerWolv:
Great thanks :)

PR #10, opened 2021-01-28
Fix : Added pragma for multiple PE MIME types.

ThisALV:
Some Windows users may have .exe MIME type which isn't application/x-dosexec.
This PR adds pragma for the two other main MIME types used by PE files.

(It was the reason for my ImHex installation to not show up the patterns popup when opening a Portable-Executable file.)

### DIFF
--- a/magic/portable_executable_magic
+++ b/magic/portable_executable_magic
@@ -1,6 +1,6 @@
 # A libmagic database containing definition for PE files used by MS-DOS based systems
 
 # MS-DOS Portable Executable
-0x0		string/b	MZ
->(0x3c.l)	string/b	PE\0\0
+0x0		    string/b	MZ      MS-DOS Binary
+>(0x3c.l)	string/b	PE\0\0  Portable Executable
 !:mime application/x-dosexec
--- a/magic/portable_executable_magic
+++ b/magic/portable_executable_magic
@@ -1,0 +1,6 @@
+# A libmagic database containing definition for PE files used by MS-DOS based systems
+
+# MS-DOS Portable Executable
+0x0		string/b	MZ
+>(0x3c.l)	string/b	PE\0\0
+!:mime application/x-dosexec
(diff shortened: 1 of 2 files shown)


WerWolv:
Hi thanks for the PR!
As a note, `application/octet-stream` usually refers to "Generic Binary Data" or data whose real type is unknown. Because of that, I don't think it's a good idea to add that here. It will cause other unknown data to be falsely identified as a PE executable as well. Instead consider creating a new magic database file that correctly identifies that file as a PE executable.

ThisALV:
I'm quite confused, should I build magic files by modifying CMakeLists.txt and putting a new magic file into magic_dbs directory ?

WerWolv:
No. First of all, try to find out why the PE executable you were loading that showed up as `application/octet-stream` instead of actually as PE executable. If it's because your file was somehow corrupted or anything, don't bother. If it's really a new type, you can make a new magic db file and PR it to this repository as well. They go into the [/magic](https://github.com/WerWolv/ImHex-Patterns/tree/master/magic) folder here.

ThisALV:
What actually happens is that the only way to get a file mime-type in ImHex is by having a .mgc file which detects the mime-type.

As magic/magic_dbs.mgc into built project only contains data about Nintendo files, basically all usual files are detected as application/octet-stream, at least when the magic_dbs.mgc file is present, elsewhere the mimeType local of ViewPattern FileLoaded event handler is just an empty string.

Now we have two solutions about this issue that can only be performed at ImHex level :
- Import the main magic database usually present into /usr/share/misc
- Modify ImHex to load mime-type with filesystem lib you're using

Note that first solution would need to include precompiled /usr/share/misc/magic.mgc into project, or to build /usr/share/misc/magic with nintendo_magic by modifying CMake configuration.

ThisALV:
Hi. Quite busy but I finally finished to fix mime-type detection adding correct magic file.

PR #12, opened 2021-03-19
Patterns for .zip files and .wav files

UserBlackBox:
Pattern for zip file parses End of Central Directory Header and Central Directory File Headers

Pattern for wav file parses RIFF header, WAVE header, and PCM header


WerWolv:
Thanks!

PR #13, opened 2021-04-20
Handles application/x-sharedlib mime-type as ELF binary

ThisALV:
ELF binaries sometimes are shared libraries rather executables. To address this issue, `application/x-sharedlib` pragma was added top of elf.hexpat file.
Also, README.md has been updated to follow current PR and previous PR, where I forgot to add Portable Executable into Magic files entries.


WerWolv:
Thank you!

PR #17, opened 2021-06-16
Pattern for SPIR-V 1.5 rev 4

rAzoR8:
This PR adds a pattern for SPIR-V https://www.khronos.org/registry/spir-v/ binary shaders.

Instructions have variable length and SPIR-V has now footer or EOS so I hardcoded 1024 instructions for now. Is there a better way to do this with builtins?


WerWolv:
That's really cool, thanks a lot for the PR!

There's a severe need for loops actual code execution in the pattern language which is needed to figure out the number of instructions there actually are. Only thing I can think of is maybe you can use the `dataSize()` builtin function together with the `$` operator to calculate it.

patterns/spirv.hexpat
```diff
@@ -0,0 +1,33 @@
+#include "spvopcodes.hexpat"
+#include "spvgenerators.hexpat"
+
+struct Version {
+    u8 dummy1;
```
WerWolv, line 5: Instead of dummy variables you can also use the `padding[N]` syntax to skip N bytes forwards in the parsing

includes/spvgenerators.hexpat
```diff
@@ -0,0 +1,33 @@
+enum GeneratorID : u16 {
```
WerWolv, line 1: Are these IDs and opcodes relevant for other data types too? If not, I'd put them into the main file. Includes are meant as sort of small libraries of types (and later on functions) that can be included and used by other people.

rAzoR8: They are not relavant for other data types as far as I know. I separated them because they get updated independantly from the SPIR-V file layout. I can merge them back if needed :)

rAzoR8: I have updated the PR & merged the enums to the spirv.hexpat :)

patterns/spirv.hexpat
rAzoR8: Ah cool, yeah thats a bit nicer! updated :)

rAzoR8:
"Only thing I can think of is maybe you can use the dataSize() builtin function together with the $ operator to calculate it."
 I dont see how, i'd still need to somehow parse all the instructions to know how many there are, so yeah loops would be nice :)

rAzoR8:
I think this PR could be merged now, or is there anything else I should fix / improove?

WerWolv:
Yeah sorry, it's looking great now.

WerWolv:
Short update, I just added while loops for arrays :)

PR #18, opened 2021-06-16
Starting to flesh out a pattern for PNG

fosterbrereton:
Please let me know if you have any feedback to improve the pattern. Thanks.


DavidBuchanan314:
IMHO it is cleaner to define the chunk type as `char type[4];`, as opposed to `u32` - then it is immediately visible what the chunk type is, and you don't need to mess around with defining numeric constants.

DavidBuchanan314:
Rather than hardcoding the number of chunks, you can keep reading until the end of a file, like so:
```
chunk_t  chunk_set[while($ < std::mem::size())];
```
This is a bit of a hack, since PNGs can occasionally have appended data, and still be considered valid by most real-world parsers.

I wrote this function to correctly count the number of chunks:

```
fn count_chunks() {
	u128 chunk_count = 1;
	u128 offset = 8;
	
	while (std::mem::read_string(offset + 4, 4) != "IEND") {
		offset = offset + 12 + (be u32(std::mem::read_unsigned(offset, 4)));
		chunk_count = chunk_count + 1;
	}
	
	return chunk_count;
};
```

However, it doesn't work correctly, I believe due to a bug in ImHex (it returns the correct count, but seems to break the parsing of chunks)

fosterbrereton:
@DavidBuchanan314 Thank you for these recommendations. Unfortunately I will not be able to see this PR through to completion, as I've been drawn away to other tools. I'll defer to the community with what to do with this PR - close it out, or transfer its ownership off to someone else.

DavidBuchanan314:
I'd be happy to pick up where you left off - I think I can do so if you give me access to your fork of the repo @fosterbrereton. Alternatively I can just open a new PR

UserBlackBox:
@fosterbrereton I tried your pattern out in a CTF forensics problem and have some improvements I would like to contribute. Can you give me access to your fork, so I can make contributions?

fosterbrereton:
@UserBlackBox Done, but please consider making a fork. For both you and @DavidBuchanan314, I'll be taking down my fork in a month or so.

UserBlackBox:
@fosterbrereton I will pull your changes into my fork then. Thanks!

PR #31, opened 2022-02-19
patterns: improve WAV pattern

PredatorCZ:
Hi,
I've improved wav pattern for a bit.


WerWolv:
These both look really great, thank you!

PR #32, opened 2022-02-19
patterns: add std::mem::find_sequence_in_range

PredatorCZ:
Implemented function from https://github.com/WerWolv/ImHex/pull/448

### DIFF
--- a/includes/std/mem.pat
+++ b/includes/std/mem.pat
@@ -22,9 +22,12 @@ namespace std::mem {
     };
 
     fn find_sequence(u128 occurrence_index, auto ... bytes) {
-        return builtin::std::mem::find_sequence(occurrence_index, bytes);
+        return builtin::std::mem::find_sequence_in_range(occurrence_index, builtin::std::mem::base_address(), builtin::std::mem::size(), bytes);
     };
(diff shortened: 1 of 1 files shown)


includes/std/mem.pat
```diff
@@ -22,9 +22,12 @@ namespace std::mem {
     };
 
     fn find_sequence(u128 occurrence_index, auto ... bytes) {
-        return builtin::std::mem::find_sequence(occurrence_index, bytes);
+        return builtin::std::mem::find_sequence_in_range(occurrence_index, 0, 0, bytes);
```
WerWolv, line 25: This function should take `std::mem::base_address()` and `std::mem::size()` in instead of just `0, 0` pretty sure.
Also don't forget to include `<std/mem.pat>` then

WerWolv (changes requested):

includes/std/mem.pat
PredatorCZ: Props to that, it would be much more readable.

WerWolv (approved):

WerWolv:
Thank you!

PR #35, opened 2022-03-23
patterns: Add GNU program types to ELF

ThePixelGamer:
https://github.com/torvalds/linux/blob/master/include/uapi/linux/elf.h#L38-L41
I'm unsure if this file format allows for exclusion of the namespace so I included it just in case


WerWolv:
I am so sorry for neglecting this for so long. Thanks a lot!

PR #38, opened 2022-07-30
Change 00=() to space bar

Berylskid:
Currently, the decode column is too wide horizontally, making it unusable. This PR is intended to solve that.


WerWolv:
Thank you!

PR #40, opened 2022-08-05
Implement VarInt type

dondish:
VarInt is commonly used in Protobuf.
It is an unsigned integer, that is dynamically sized.

In order to parse a VarInt, you iterate over each byte, and if the msb of the byte is 1, then it is not the last byte. The least significant 7 bits are then prepended to the binary representation of the variable.

For example, let's take the following 2 bytes:
`1010 1100 0000 0010`
The msb of the first byte is set, while for the second one it is not, so the length of the VarInt is 2 bytes.
We then take the 7 least significant bits of the two bytes, and prepend the latter to the former:
`000 0010 ++ 010 1100`
We get `1 0010 1100`, or 300 in decimal.


WerWolv:
Hi
Thanks a lot for this addition!
I think this type is usually referred to as LEB128, or is this something else? We have support for it in the ImHex data inspector so it would be nice if it was named the same here

dondish:
It is the same, I updated the names.

WerWolv:
Thank you!

PR #41, opened 2022-08-08
Add enhancements for pe.hexpat

gmestanley:
The current pattern file for the PE format doesn't have a lot of the format's quirks, so I decided to code them in after I noticed that it doesn't cover the structure known as Rich Header. (Forgive the garbage code for its ProductType enum, it was the only way I found to make the values appear.)
I changed the MIME type to x-msdownload so that it supports PE file extensions other than .exe.
Here are my sources for the improvements included here:

How the MZ header works and some of its variables' names: [How to determine the size of an PE executable file from headers and or footers](https://stackoverflow.com/questions/34684660/how-to-determine-the-size-of-an-pe-executable-file-from-headers-and-or-footers)
The function of some of the MZ header variables: [PE102 by Corkami](https://github.com/corkami/pics/blob/master/binary/pe102.png)
The existence of sections: [PE101 by Corkami](https://github.com/corkami/pics/blob/master/binary/pe101/pe101.png)
The Machine values for LoongArch processors, the architecture enum and how it's used in the Optional Header, Subsystem types, DLL & Section characteristics, how sections, their line numbers and relocations work: [PE Format](https://docs.microsoft.com/en-us/windows/win32/debug/pe-format#the-rsrc-section)
The Machine values for DECAlphaAXP and i860: [Peering Inside the PE: A Tour of the Win32 Portable Executable File Format](https://docs.microsoft.com/en-us/previous-versions/ms809762(v=msdn.10)#the-pe-header)
How the Rich Header works: [VB2019 paper: Rich Headers: leveraging this mysterious artifact of the PE format](https://www.virusbulletin.com/virusbulletin/2020/01/vb2019-paper-rich-headers-leveraging-mysterious-artifact-pe-format/)
Values of products in the Rich Header: [The devil’s in the Rich header](https://securelist.com/the-devils-in-the-rich-header/84348/)
Every other value not found in the above source: Ghidra

### DIFF
--- a/patterns/pe.hexpat
+++ b/patterns/pe.hexpat
@@ -1,4 +1,10 @@
-#pragma MIME application/x-dosexec
+#pragma MIME application/x-msdownload
+#pragma pattern_limit 100000
+#pragma array_limit 10000
+#include <std/mem.pat>
+#include <std/string.pat>
+#include <std/ctype.pat>
+#include <std/io.pat>
 
 enum MachineType : u16 {
 	Unknown = 0x00,
@@ -7,9 +13,13 @@ enum MachineType : u16 {
 	ARM = 0x1C0,
 	ARM64 = 0xAA64,
 	ARMNT = 0x1C4,
+	DECAlphaAXP = 0x183,
 	EBC = 0xEBC,
 	I386 = 0x14C,
+	I860 = 0x14D,
 	IA64 = 0x200,
+	LOONGARCH32 = 0x6232,
+	LOONGARCH64 = 0x6264,
 	M32R = 0x9041,
 	MIPS16 = 0x266,
 	MIPSFPU = 0x366,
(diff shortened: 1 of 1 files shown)


WerWolv:
Hi
Thanks a lot for the effort, that's really cool!

Currently the Unit Tests are failing however when it tries to run your pattern against the test exe file: 
image
Could you investigate this? The test file can be found here: 
https://github.com/WerWolv/ImHex-Patterns/tree/master/tests/patterns/test_data

gmestanley:
Can do. Sorry for the late reply, will try investigating it now

gmestanley:
I can't seem to find any problem with the code, it runs just fine with the test executable you provided (shown in the image below).
portableexecutable
I made a version of the .hexpat file with different positioning of the Section struct, can you try it and see if the Test Units work with that?
pe.hexpat

WerWolv:
Turns out this was a bug in the pattern language itself, I fixed it now. Sorry for blaming you for it.

Both the old and the new pattern seem to work fine now. Did you want to have the pattern with your latest changes merged or the one that has been PRed here now?

WerWolv:
Okay great. Thanks a lot!

PR #44, opened 2022-08-19
WAV: Fix related to chunk size.

jz5:
* Consider appended padding, use chunk size.
* Fix to add 1 byte length for padding if chunk size is an odd number of bytes.


WerWolv:
Thanks a lot for the fixes!

PR #46, opened 2022-09-02
encodings: Add euc_kr table

mirusu400:
I sorted euc_kr table as lexicographic order. If we need to sort as hex order, please comment here.


WerWolv:
Thanks a lot! 
All the other encodings are sorted in numerical order instead of lexicographical order. While it doesn't matter to ImHex, I think it would be nice to have all the files the same.
Could you also add a new entry for your encoding to the readme?

mirusu400:
Edited 😉

WerWolv:
Thanks a lot :)

PR #48, opened 2022-10-08
Add pattern for ID3 tags

flimberger:
I found the patterns for OGG and Vorbis useful when developing a custom metadata tag parser, but there was none for ID3, so I created one.

It's still missing support for extended header data, but I have no such files in my library.


WerWolv:
Hi thanks a lot for the pattern!

Please also add your new pattern to the list of patterns in the readme and set a MIME type for your pattern using `#pragma MIME your/type` if it has one. Also if you have a small test file, you can upload it to `/tests/patterns/test_data` so your pattern gets verified using the unit tests

WerWolv:
Thank you!

PR #50, opened 2022-10-11
Fixed: Elf pattern 32 bit

fr0ntlin3:
Added structure to the body, renamed data_size -> size, works fine now


WerWolv:
Thanks for the fix! Did you mean to delete your Fork and close the PR? Otherwise I can add the fix myself

PR #52, opened 2022-10-30
Making pe.hexpat even better

gmestanley:
Hi! Bringing in a ton of additions, including:
 - Code for the Resource/.rsrc section has all the values other than the header now;
 - Statement `Product products[while($ != richHeaderEndPosition)];` in the code for the Rich Header was changed to use `<` instead;
 - Debug and CRT sections were implemented;
 - The code declaring sections was touched up;
 - Code for the DOS stub is now more accurate;
 And a few other things that were fixed, such as the Read-only data Section's code. I know that it doesn't work that way as do sections in general, but trying to implement that took too long. This is only for publishing the fixes that have been already made.

### DIFF
--- a/patterns/pe.hexpat
+++ b/patterns/pe.hexpat
@@ -32,7 +32,7 @@ u16 dosMessageOffset;
 fn isntdosmessage(char c) {
 	return c == 0x0D || c == '$';
 };
-  
+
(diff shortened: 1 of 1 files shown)


WerWolv:
That is really really cool. Thanks a lot!

PR #53, opened 2022-11-04
patterns: png.hexpat: APNG support

ctrlcctrlv:


patterns/png.hexpat
```diff
@@ -1,17 +1,21 @@
 #pragma MIME image/png
 #pragma endian big
+#pragma pattern_limit 0
```
WerWolv, line 3: What's the reason this is needed? Since this can lead to crashes due to ImHex running out of memory on less well equipped computers, this should usually just be added by the user if needed.

WerWolv:
Thanks a lot for the improvements!

PR #56, opened 2022-11-07
patterns: fix java class pattern and update to javase17 format

lieyanqzu:
1. Add new constant type and access flag
2. Fix constant pool length error
3. Fix variable name wrong
4. Add one test class file


WerWolv:
Thanks a lot!

PR #57, opened 2022-11-10
Add DOS datetime types to pattern library

kavika13:
This matches the implementation in the inspector for DOS Date and DOS Time. Useful for old formats (such as ZIP).


nicolas17:
Why was this merged into dev rather than master? It's not available in the store :thinking:

PR #58, opened 2022-11-12
Adding pattern file for NE executables

gmestanley:
Started this back in July, only got to making a presentable version now - sorry for the wait!
Sources: [BYTE* / Windows 3.0 16-bit NE Executable Format](http://bytepointer.com/resources/win16_ne_exe_format_win3.0.htm), [NE - OSDev Wiki](https://wiki.osdev.org/NE)


gmestanley:
Earlier when I submitted this pull request, the Unit Tests reported a code failure; it actually is one and in fact a recent mistake. I was led to believe the variable EntryTablePointer is a pointer by the first source, so I renamed it from EntryTableOffset but didn't update the references to it in the code, sorry about that. Turns out however that treating it like a pointer is wrong. because it assigns the entryTable pattern to an unrelated address and on top of another pattern. Actually treating it like an offset as the second source says puts it in a more appropriate place, so I did that instead.

WerWolv:
Thanks a lot! That's really cool

Please also add a link to the new pattern to the list in the readme

gmestanley:
Did that now.

WerWolv:
Thanks!

PR #59, opened 2022-11-14
Minor cleanup to zip pattern

nicolas17:
I did some minor cosmetic changes to zip.hexpat: cleanup whitespace, remove an unused struct, and shorten a giant field comment that was causing a tooltip wider than my screen.

I'm sending this PR to keep these minor changes separate from my actual additions later :)

By the way, I see most commit messages start with the directory/category name like "patterns:". I did the same for consistency, but if only touching one file, wouldn't it make more sense to put the format name there? Like "png: Add sRGB chunk support".

### DIFF
--- a/patterns/zip.hexpat
+++ b/patterns/zip.hexpat
@@ -3,44 +3,44 @@
 #include <std/mem.pat>
 
 struct EndOfCentralDirectory{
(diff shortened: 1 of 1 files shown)


WerWolv:
Thank you!

PR #60, opened 2022-11-15
zip: Change how the end of central directory record is found

nicolas17:
Previously, the zip pattern was searching for the end-of-central-directory header signature (50 4B 05 06) by searching the entire file for it. This is *very* slow for large files, and risks false positives since those bytes could randomly appear in compressed data. I had this happen on the first large (>2GB) zip file I tried.

I'm now checking for the EOCD signature at exactly 22 bytes from the end of the file (in the common case there is no zip comment), and if that fails I search for it in the last 64KB of the file (in case there *is* a comment at the end of the EOCD, which can't be larger than 64KB). This is much faster, and fixes loading my zip file where it was spuriously finding the signature in the wrong place.

This still has a low risk of false positives (what if the comment has the 50 4B 05 06 bytes? what if there is a short comment but the signature appears in the last 64KB of compressed data?), but I don't know what's the "right" way to find the EOCD to avoid those edge cases, or how proper zip-reading tools do it...

### DIFF
--- a/patterns/zip.hexpat
+++ b/patterns/zip.hexpat
@@ -1,6 +1,7 @@
 #pragma MIME application/zip
 
 #include <std/mem.pat>
+#include <std/math.pat>
 
 struct EndOfCentralDirectory {
     u32 headerSignature [[color("00000000")]];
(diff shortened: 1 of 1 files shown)


WerWolv:
Thank you!

PR #63, opened 2022-12-02
Add pattern file for Apple PBZX compressed streams

nicolas17:
This is used by Apple for Xcode .xip files (cpio in pbzx in xar), OTA updates of iOS, and other files.


WerWolv:
Thank you!

PR #65, opened 2022-12-19
Enhancing shiftjis.tbl

gmestanley:
shiftjis.tbl had an annoying problem where null bytes appeared as (), which clutters the ASCII (technically ANSI?) view. I solved the problem by removing the 00= part, and also changed the other 0x0X characters to not appear as strings that would definitely also clutter the view.


WerWolv:
Thank you!

PR #66, opened 2022-12-27
includes/std: Fix namespacing for std::limits::s*_max()

123456abcdef:
Added missing namespace to `std::limits::s*_max()`.


WerWolv:
Thank you!

PR #70, opened 2023-01-08
patterns/dmg: Added DMG pattern

shellcromancer:
# Pattern: DMG

This adds the struct/pattern for Apple Disk Images's trailer object (.dmg). This could be further extended by decoding the XML identified as `metadata_plist`, decoding the base64 payload within that and parsing those structs and the compressed objects they relate to if base64 & XML could be added to the[ ImHex decoding library](https://imhex.werwolv.net/docs/libraries/hex/hex_dec.html#decoding-library-1-24-0). 🤔

image

## Checklist


WerWolv:
Very cool thanks a lot! :)

I'm certainly open to add more decoding options to the decoding library

WerWolv:
Oh quick note that I should probably add to the issue template.
`application/octet-stream` shouldn't be used as the MIME type as it means "raw data without any identifiable format". So your pattern will be suggested for every unknown binary.
I'll fix it

PR #73, opened 2023-01-11
Update shiftjis.tbl and create ms932.tbl

Berylskid:
- Duplicated the contents of shiftjis.tbl as ms932.tbl before the change.
- Resolved some character code confusion in shiftjis.tbl.


Berylskid (commented):
Add a missing character (819F=◆)

WerWolv:
Thank you!

PR #76, opened 2023-01-16
Add PIF image pattern

gfcwfzkm:
PIF pattern with a test image for ImHex

More about the pif image format: https://github.com/gfcwfzkm/PIF-Image-Format

Thanks for this great hex viewer!


WerWolv:
Thank you!

PR #77, opened 2023-01-17
patterns/tar: support PAX archives

hdonnay:
There are two common magic values for archives: the [GNU one](https://git.savannah.gnu.org/cgit/tar.git/tree/src/tar.h#n160) (8 bytes) and the [ustar one](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/pax.html#tag_20_92_13_06) (6 bytes).

This change has the pattern support both.

### DIFF
--- a/patterns/tar.hexpat
+++ b/patterns/tar.hexpat
@@ -9,7 +9,8 @@
 #include <std/sys.pat>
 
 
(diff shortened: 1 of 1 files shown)


WerWolv:
Thank you!

PR #78, opened 2023-01-31
Support null-terminated ustar magic value

willemv:
The GNU tar cli tool creates tar files where the magic "ustar " constant ends with a space instead of a nul-byte.

The [spec](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/pax.html#tagtcjh_28) says it should be a nul-byte though, so just check the first 5 characters and be flexible about the 6th :)

### DIFF
--- a/patterns/tar.hexpat
+++ b/patterns/tar.hexpat
@@ -9,7 +9,7 @@
 #include <std/sys.pat>
 
 
-#define GNU_HEADER "ustar "
+#define GNU_HEADER "ustar"
 #define PAX_HEADER "ustar\x00"
 #define NULL "\x00"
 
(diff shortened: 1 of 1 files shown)


KillyMXI:
How is this different from the existing code that already checks against both space and null byte?

WerWolv:
Hi, thanks for the PR!
You should probably also adjust the define name and the assert right below. `magic == PAX_HEADER` will now always be false and the `GNU_HEADER` and `PAX_HEADER` names make no sense anymore since the only thing identifying the difference there is the last byte. So probably checking for a value named something like `TAR_HEADER_MAGIC` would make more sense now

willemv:
My bad, I was still working with and old version of ImHex (1.26.2) where this bug wasn't fixed yet, and simply fired of this PR without checking whether it was fixed already.

I'll close this PR since it isn't useful anymore.

Sorry to bother you :blush:

PR #79, opened 2023-02-03
Added missing int array and long array tag

Lenni0451:
The pattern code for the IntArray and LongArray tags were missing in the NBT.hexpad


WerWolv:
Thanks a lot!

PR #83, opened 2023-02-12
Fix the IPv6 type

Kuruyia:
This fixes the IPv6 type by importing the string library for the format function, and forcing the words to have the network order.

Leading zeros have also been removed from the formatted string, because they are not necessary when displaying in short form.


WerWolv:
Thank you!

PR #86, opened 2023-02-17
Add signed LEB128 type support

Chrisg2000:
Add support for signed LEB128 type in the standard type library.

Also added documentation to the Pattern Language Docs: WerWolv/PatternLanguage#26


WerWolv:
Awesome work! Thanks a lot

PR #92, opened 2023-02-26
ELF: make hex view less ambiguous

amaksoft:
**DO NOT MERGE BEFORE THE CORRESPONDING PATTERN LANGUAGE PULL REQUEST**
https://github.com/WerWolv/PatternLanguage/pull/29

## Problem statement:

Since ELF file has ~30 sections and ~8 segments all with similar structures, it gets very ambiguous in Hex Editor view:

Screenshot 2023-02-26 at 17 15 29

What section/segment do each of `p_data` and `data` belong to?
To figure it out the ImHex user ends up jumping back and forth between Hex Editor and Pattern Data views and clicking to expand each section/segment header as they are not highlighted when you highlight some byte in Hex Editor view.

## Proposed Solution:

We add a post processing step to set the display names for all data data arrays to the notations that mention both section index and name, index and type for segments. Unfortunately it has to be a post processing step as string table with all of the names is also one of the sections. Not even close to being the first one.
Result:

hover_over_hex_ELF

P.S. Sorry if my pattern script is a bit clunky, i am ready to improve it with your guidance :)

P.P.S A more complicated alternative to it will be making `pl::ptrn::Pattern` store the reference to its parent and show the path from pattern node to root pattern in pattern tooltip view, but it will be less flexible.

### DIFF
--- a/includes/std/core.pat
+++ b/includes/std/core.pat
@@ -57,4 +57,13 @@ namespace std::core {
     fn is_valid_enum(ref auto pattern) {
         return builtin::std::core::is_valid_enum(pattern);
     };
+
+
+    fn set_pattern_color(ref auto pattern, u32 color) {
+        builtin::std::core::set_pattern_color(pattern, color);
+    };
+
+    fn set_display_name(ref auto pattern, str name) {
+        builtin::std::core::set_display_name(pattern, name);
+    }; 
 }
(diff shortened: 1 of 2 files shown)


WerWolv:
I merged your PatternLanguage PR so this can be merged once a new version of ImHex gets released. Thank you :)

One nit-pick, you should generally never use the functions from the `builtin::` namespace directly. Instead, since you added `builtin::std::core::set_dislpay_name` now, add a function with the same name to the standard library in the `includes/std/core.pat` file please

PR #94, opened 2023-03-09
Adding Dalvik Executable pattern

cryptax:
It's not exactly the entire DEX format (https://source.android.com/docs/core/runtime/dex-format) but personally, I don't need more than this.

Uses the ULEB128 types -- possibly need to update Content Store to get the pattern work.


WerWolv:
Thanks a lot! :)

PR #95, opened 2023-03-11
add more information to display for dex pattern

eybisi:
Added string representations for:
- string ids
- field ids
- method ids
- class ids
- proto ids

Example display:

Screenshot_85

I left the `class_data_item *class_data_off:u32;` line commented out because it increases parsing time so much. In my test dex sample 0.5 to 3.5 seconds


WerWolv:
Thank you!

PR #97, opened 2023-03-18
Add nullable pointers to std/ptr

ammkrn:
An implementation of nullable pointers was discussed briefly on discord; WerWolv suggested the implementation in the example code here that's ostensibly generic over the pointer type/size, but the template system doesn't seem to like this special case where a type argument is used as a pointer type (the code below returns the error `Pointer size needs to be a built-in type`, with the error being directed at `T *data : Size;`):

```
struct NullablePtr<T, Size> {
    Size pointerValue [[no_unique_address, hidden]];
    if (pointerValue == 0x0) {
        padding[sizeof(Size)];
    } else {
        T *data : Size;
    }
};
```

So I specialized the structure to the common 2^n pointer types. I'll leave it up to WerWolv /  maintainers whether they want to use this specialized version, do something else, or just forego adding this as a feature and instead document it as a "user pattern" in the manual or something.

### DIFF
--- a/includes/std/ptr.pat
+++ b/includes/std/ptr.pat
@@ -34,4 +34,34 @@ namespace std::ptr {
         return std::mem::size() - offset * 2;
     };
 
+    /**
+    A nullable pointer, generic over both the pointee type and pointer type.
+
+    By nullable, we mean that if the pointer's value is zero (`0x0`), then the
+    value will appear as padding rather than a pointer to something, but
+    if the pointer's value is non-zero, that non-zero value will be treated as
+    a pointer of type `PointerTy` which points to an element of type `PointeeTy`.
+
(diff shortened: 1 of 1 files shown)


WerWolv:
Thanks a lot!
I just added a new feature in the latest ImHex commit that allows any types to be used as pointer types, including completely custom types and also template parameters. If you want to revisit your PR again, your initial goal is now possible :)

ammkrn:
I've updated it to reflect the change; the generic implementation in 96fca8c works as of the nightly generated by [6a9f796](https://github.com/WerWolv/ImHex/commit/6a9f79628e53c9efab852033e4628657289f88ba)

WerWolv:
Thank you!

PR #98, opened 2023-03-18
Update the pbzx pattern file

nicolas17:
- Rename pbzx to pbz (including the pattern filename); turns out the 'x' is the compression type.
- Use type::Magic for the "pbz" magic number.
- Decode compression type as an enum.
- Mention compression_tool in the header comment.


WerWolv:
Thank you!

PR #99, opened 2023-03-23
Patterns/PCAP: improve the pattern

puhitaku:
Hi, I've made some changes to pcap.hexpat:
 - Reformat the lines to use spaces instead of tabs
 - Switch endianness by the magic number's endianness
 - Consume all packets until the EOF

Thank you for developing ImHex!


WerWolv:
Thanks a lot!

PR #100, opened 2023-03-25
add evtx pattern

doracpphp:
Created a header and chunk pattern for MS EventLog in Windows XML Event Log (EVTX) format.

Windows XML Event Log (EVTX) format Document 
https://github.com/libyal/libevtx/blob/main/documentation/Windows%20XML%20Event%20Log%20(EVTX).asciidoc


WerWolv:
Thank you!

PR #106, opened 2023-04-17
Recently added exp function needs to be added to math include

paxcut:
Removed extraneous argument base to ln function and added the necessary definition in include file to access previously builtin function exp.


WerWolv:
Thank you!

PR #110, opened 2023-04-24
Added Crash Bandicoot - Back in time level format

AdventureT:
User created flashback tapes level format pattern.
From the fan game Crash Bandicoot - Back in Time


WerWolv:
Thanks a lot!

PR #112, opened 2023-05-16
Update pe.hexpat

gmestanley:
Hello! Here I am again with the new version of pe.hexpat that I promised.
I've been fiddling around with it since I gave the last one and now it's finally ready.
Here's a general gist of the changes:

Most variables' names were made clearer;
Better formatting;
Renamed checkForDebugInformation to checkForSymbols (clearer about what it does and fits the pattern editor tab better);
Replaced the whole section system with a new one that more accurately detects structures inside files and can detect more than only one of them per section;
Redid the Imports table code (older one was too linear and doesn't work on some files);
Fixed Dialog resource in the Resource table assigning integer values to obvious strings;
Added new machine types and Characteristics values (source is [Wikibooks](https://en.wikibooks.org/wiki/X86_Disassembly/Windows_Executable_Files)), Rich Header product values (source is Ghidra) and DLL Characteristics values (source is old Microsoft documentation on PE ["Peering Inside the PE: A Tour of the Win32 Portable Executable File Format"](https://docs.microsoft.com/en-us/previous-versions/ms809762(v=msdn.10)#the-pe-header))

### DIFF
--- a/patterns/pe.hexpat
+++ b/patterns/pe.hexpat
@@ -1,10 +1,101 @@
 #pragma MIME application/x-dosexec
 #pragma MIME application/x-msdownload
-#pragma pattern_limit 0x400000
 
 #include <std/string.pat>
 #include <std/mem.pat>
 
+// Rich Header
+bool checkForRichHeader in;
+u16 richHeaderEndPosition;
+fn initializeRichHeader() {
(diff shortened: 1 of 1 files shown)


WerWolv:
Thanks a lot!

PR #114, opened 2023-05-25
Improved GIF pattern

blondecake433:
Hey :)
I was trying to fix a malformed GIF and found the pattern to be a little lacking so I improved it a bit.

This PR contains the following:
- Changed some structs names, and the way they are parsed to be consistent with the [89a specification.](https://www.w3.org/Graphics/GIF/spec-gif89a.txt)
- Removed function implementations that were already in the standard library.
- Added a MAGIC assert and some warnings when unexpected patterns are found.
- Removed un-helpful `[[comment]]` and `[[hidden]]` tags.
- Cleaned the code from commented TODOs and commented out lines.

### DIFF
--- a/patterns/gif.hexpat
+++ b/patterns/gif.hexpat
@@ -1,20 +1,23 @@
+#pragma MIME image/gif
+
+// Extension Labels
 #define LABEL_GC 0xF9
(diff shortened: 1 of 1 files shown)


WerWolv:
Thank you!

PR #115, opened 2023-06-01
Fix macho CpuType error and add CpuType::ARM64 and CpuType::ARM64_32

cfeitong:
Refer to LLVM's [document](https://llvm.org/doxygen/namespacellvm_1_1MachO.html#af2acbc063e449084ba33f738a700ce47)


WerWolv:
Thank you!

PR #117, opened 2023-06-04
Added visualizers to image patterns

paxcut:
There were two patterns that couldnt have the visualizer integrated because they don't use a completed final pattern.


WerWolv:
Thank you!

PR #122, opened 2023-06-06
pattern/java_class: Format byte code instruction

Guigumua:
Reference: https://docs.oracle.com/javase/specs/jvms/se20/jvms20.pdf


Nowilltolife:
Would't this also interpret operands as instructions, making it a bit more harder to distinguish where instructions are positioned?

Guigumua:
> Would't this also interpret operands as instructions, making it a bit more harder to distinguish where instructions are positioned?

You are right, I made a mistake, I will modify the code later

PR #123, opened 2023-06-07
pattern/java_class: Format byte code instruction

Guigumua:
Format bytecode instructions into a human-readable format.

image

But there are some problems:
1. code_length describes the byte length of the instructions, but the instruction is variable length, so the length of the instruction array will not match the `code_length`.
2. Lines 1036-1047 use `addressof(parent)`, and what is expected is the starting address of the `code` array in `attribute_code`, but it is actually 8 bytes less.

It works, but possible to optimize.

### DIFF
--- a/patterns/java_class.hexpat
+++ b/patterns/java_class.hexpat
@@ -141,6 +141,234 @@ namespace fmt {
             (_): return fmt::const_ref(index);
         }
     };
+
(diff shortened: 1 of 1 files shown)


WerWolv:
Thanks a lot!
I'll definitely look into those problems

PR #124, opened 2023-06-11
pattern/midi: Support multiple tracks

qufb:
Expands on the existing pattern, allowing to parse files regardless of how many tracks or types of events they contain.

Also included a public domain test file: https://www.mutopiaproject.org/cgibin/piece-info.cgi?id=931

Screenshot from 2023-06-11 10-29-45


WerWolv:
Thanks a lot!

PR #127, opened 2023-06-15
Fix pattern to see data section of DEX files + Link_data

cryptax:
Rather than a Fix, this is more additional support to see more inside the DEX file format.


cryptax:
Unit Test fail looks like CI issue, but not related to this MR.

WerWolv:
Thanks! Yeah the CI has issues right now because it can't upload coverage data for some reason. Not your fault.

PR #129, opened 2023-06-17
Added mp4 pattern

blondecake433:
Hello :)

I created a mp4 hex pattern.
It is not complete as this format is very [extensive](https://b.goeswhere.com/ISO_IEC_14496-12_2015.pdf), But I made this pattern with this in mind, 
so as it won't fail when parsing unknown data types but simply indicate it with a special data type `UnknownBox`.


WerWolv:
Hi thanks a lot for your work! 
Many patterns only parse parts of a format, especially with formats as complex as MP4 so that's absolutely no concern :)

PR #138, opened 2023-07-10
patterns/bmp: various improvements

brliron:
- Add various version of the BitmapInfoHeader struct, and autodetect the correct one.
- Fix palette detection.
- Use BitmapInfoHeader::biSize to find the position of the palette / data.


WerWolv:
Thanks a lot!

PR #142, opened 2023-07-16
Add PyInstaller's pattern

lopqto:
Greetings,

In this pull request, I have implemented a pattern to parse PyInstaller binary headers. PyInstaller is a tool that packages a Python application along with its dependencies into a single executable, eliminating the need for a separate Python interpreter or modules installation by the end user.

I am also familiar with the Contribution Policy and the requirement to include a test file. However, I would like to address the issue regarding the test file's size limitation. Due to the nature of PyInstaller's binaries, it is not feasible to create a test file that meets the specified requirement of being 100 kilobytes or less.

For a standalone executable, such as a simple "hello world" application, the file size is approximately 1.5 megabytes:

```bash
$ du -h dist/helloworld/helloworld
1.4M	dist/helloworld/helloworld
```

On the other hand, for a bundled executable, the file size starts at around 7 megabytes. Given this significant difference in file sizes, I kindly request confirmation regarding the limitation on test file size. If you acknowledge this constraint, I will be more than willing to add a test file accordingly.

image


WerWolv:
Hi, thanks a lot!
A 1.5MB file is okay as an exception, the rule mostly stands so people try to get their example files as small as possible to keep the repo small and fast to clone.
If you can, truncate the the executable to only include data that's actually being parsed by your pattern. For example, if your pattern only parses the header data but none of the actual executable data, it would make sense to only include that header.

lopqto:
Thank you for your prompt reply.
I managed to remove all the unnecessary modules from the binary, making it much smaller at only 71 kilobytes. It still works fine, but it needs the "libpython3.11.so.1.0" and "libz.so.1" libraries to be present on the running directory. Please, take a look at the latest commit to see the changes I made.

WerWolv:
That's pefect, thank you! 
We don't really need the executable to run still, it's just used to have the CI automatically run unit tests on them to verify that changes to the pattern language runtime didn't break any of the submitted patterns.

I'm letting the CI run on it right now so I'll just wait for that to finish and merge your PR then :)

PR #145, opened 2023-08-02
ubiquiti.hexpat?

MatrixEditor:
I wrote a pattern file for Ubiquiti firmwares. Do we need such a pattern or firmware patterns in general? Firmware can be downloaded without any authorization from [https://fw-update.ubnt.com/api/firmware](https://fw-update.ubnt.com/api/firmware), so test files may be added as well.

Since those firmware files are not smaller than ~4Mb, I can try to create a custom crafted file with some stubbed entries that can be used for unit testing.


WerWolv:
Hi, thanks a lot!
As long as the pattern is useful for other people too, even very specific patterns are appreciated :) 

If you can make a smaller test file that would be great, thank you!

WerWolv:
Thanks a lot!

PR #146, opened 2023-08-03
Fix DebugDirectory version field types

lzybkr:
majorVersion/minorVersion fields in the DebugDirectory are incorrect, they are actually u16.


WerWolv:
Thank you!

PR #153, opened 2023-08-26
Adding IPS & correction to PE magic

gmestanley:
Hello once again! I added a pattern file for the IPS file format this time.
Also included is a ips.hexpat.ips test file, and it was also added to the readme.
Source for the pattern file is: http://justsolve.archiveteam.org/wiki/IPS_(binary_patch_format)
The test file is an IPS patch file that patches ips.hexpat itself to do different logic for finding the EOF's offset (from the file's size) and adds four spaces at the end, to test RLE patching.

I also clarified the portable_executable_magic file in magic/ to mention Windows in the description, since most PE files made nowadays do not take MS-DOS in mind.


WerWolv:
Thank you!

PR #156, opened 2023-09-01
BSON support for multiple documents per file

ershov:
BSON files can contain consecutive documents glued one after another. An example of these is MongoDB FTDC metrics export.

[`bsondump`](https://github.com/mongodb/mongo-tools/blob/master/bsondump/bsondump.go) can unpack this type of BSON documents.


WerWolv:
Thanks a lot!

PR #157, opened 2023-09-02
Fix for issue 1279

paxcut:
Added the missing part at the end of the string after being replaced.


WerWolv:
Thanks a lot!

bgilbert:
It looks like this fix wasn't correct; xref https://github.com/WerWolv/ImHex/issues/1279#issuecomment-1712950402.

PR #160, opened 2023-09-09
patterns/id3: Fix the error `No variable named 'n' found.`

dn-ln:
### OS and Editor Version
- macOS Ventura 13.2
- ImHex Hex Editor v1.30.1

### Description
The console showed the error message `No variable named 'n' found.` when I opened an MP3 file with ID3 v2.4.0 tag and imported the ID3 pattern.

After tweaking the pattern, I found two ways to resolve the error:
- Remove the `ref` keyword from the argument `n` for the function `ssi_value`
- Use `size` as the variable name inside the function `ssi_value` (I changed the argument name to `size` too but it can be any name or even left empty)

I don't know which way is preferable. Maybe the problem is in the core library so the pattern file should not be changed at all. I have little knowledge about the behavior so I just opened the PR to see if it solves anything. Thanks for the good work and I enjoy the editor very much!

#### The error
id3_error
#### Remove `ref` keyword
id3_error_solution_1
#### Use `size` as the variable name
id3_error_solution_2

### DIFF
--- a/patterns/id3.hexpat
+++ b/patterns/id3.hexpat
@@ -26,10 +26,10 @@ namespace v2 {
         u8 bytes[4];
     } [[sealed, static, format("v2::ssi_value"), transform("v2::ssi_value")]];
 
-    fn ssi_value(ref SyncSafeInt n) {
+    fn ssi_value(ref SyncSafeInt size) {
         u32 result = 0;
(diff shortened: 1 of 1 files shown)


WerWolv:
Thank you!
I think this issue has been fixed since internally but I'll still merge your PR as the name of that parameter wasn't great

PR #159, opened 2023-09-03
patterns/tiff: support BigTIFF and tiled TIFF; bugfixes and cleanups

bgilbert:
Major changes:

- Support BigTIFF (64-bit TIFF)
- Support tiled TIFFs
- Clean up strip array generation; don't omit any partial last strip
- Correctly distinguish between values and offsets based on total size, rather than guessing from the field type
- Correctly represent field widths for small values in little-endian TIFFs
- Some display improvements
- Various code cleanups
- Add additional test cases

See individual commits for details.

cc: the original authors, @jacksonbarreto and @joelalves

PR #161, opened 2023-09-09
tests/patterns: support multiple test files for a pattern

### LINKED ISSUE #158: Consider supporting multiple test cases per pattern
Some patterns have to make decisions based on file headers, such as selecting a particular byte order or field width.  In such cases it's impossible to create a single test case that fully exercises the pattern.  Is there any reason the test suite shouldn't allow multiple test cases per pattern?

bgilbert:
Allow either a single test file in `tests/patterns/${pattern}.hexpat.${ext}`, as we do now, or multiple files in `tests/patterns/${pattern}.hexpat/*`.

Closes #158.  Split out from #159.


WerWolv:
Thanks a lot!

PR #162, opened 2023-09-10
patterns: add DICOM

bgilbert:
DICOM files specify a "transfer syntax" which defines the image compression format, the encoding rules for the entire file, and any special-case encoding rules for particular fields.  Many transfer syntaxes have common design features, but a transfer syntax is allowed to use completely custom encoding rules, and need not be publicly documented.

Explicitly reject transfer syntaxes we know we don't support: implicit VR (uncommon), explicit VR big-endian (deprecated), deflate-compressed. In all other cases, follow some general rules and hope it works out.

The large lookup functions were generated by a script, but I haven't included the code here, since its input is hand-massaged data copied from HTML tables.

Add a JPEG WSI test case from the WG26 2021 hackathon (CC0 license) and a synthetic LE Explicit VLP case produced with `img2dcm`.


bgilbert:
Rebased and marked ready for review!

WerWolv:
Hi, thanks a lot for your PR! Sorry for not coming back to it earlier

I know this is currently a bit stupid but could you move the constants you defined in a library back into your pattern? Otherwise people will need to download that library that's very specific to that one file before they can use the pattern itself. We'll need to find a better solution at some point to allow for easy use of multiple files per pattern but we're not there yet right now. 

Everything else looks great, I'll merge your PR if you could copy the data back into the pattern

bgilbert:
Okay, done!

PR #166, opened 2023-09-22
Add .nes pattern file

gmestanley:
Hello, I'm here with a pattern file for the .nes file format. This is the one initially conceived as the iNES format by Marat Fayzullin for files with NES game roms.

Credits go to 100th_Coin for telling me about where the entry point is located and Gilde for transcribing a kanji that described the complementaryChecksum variable.


WerWolv:
Thank you!

PR #172, opened 2023-09-29
patterns/zip: add parsing local file header with minor adjustment

HalfInner:
Introduce parsing of local file headers.
Introduce compression method header support.
Make minor adjustments: remove black colors and mark magic IDs.


HalfInner:
Lately I studied a bit a zip format, and I found out that it's very black, and LCF is not parsed, so I added support there.

image

WerWolv:
Thanks a lot!

PR #174, opened 2023-10-14
Better code + credit given for ips.hexpat & better readme

gmestanley:
Hey, new pull request here with three new things; I fixed the indenting of the PP pattern file's description on the readme looking wrong, and added x-msdownload to the list of MIME types in the PE pattern file. I also added a new version of ips.hexpat!
This adds my author stat to it which was missing in iTrooz's commit, and also makes the code quality of the file overall better.
Also a minor correction to the indenting on the preprocessor statements.


gmestanley:
It turns out if you make a modification to a fork while a pull request you made for it is being accepted, that's also included in the list. I did not know that! So I guess here's a comment on the new gb.hexpat features also.
First off I did some grammar fixes (the word cartridge was spelled as Cartrige; same with interrupts missing an R and global missing the B).
I also corrected sgbFlag and japanOnly from bools to enum values; they're not bools, if sgbFlag is anything but 3 (which would be valid `true` values besides 0) the Super GameBoy does not respond. japanOnly is also more of a destination code than a bool; if it's 0 it's for Japanese games, and if it's 1 it's for everyone else, _not_ 1 if it's a Japanese game. jmpInstruction was also corrected to an array of u8s as it's not a little-endian 32-bit value (and renamed to entryCode although I genuinely did not remember it being anything but that).
Going off of that, variables were also given clearer names, like version into maskROMVersionNumber. I added my own comments to CGB in case people can't figure out what a CGB is, and since it's not an enum and instead a bitfield, I made a compromise and made the whole bitfield have the value of NO_CGB if it's all zeroed. Finally I also added the option to have uppercase and prefixed names for the ROM features, as there's a source that lists them like that.
My sources for all of these improvements were the [Pan Docs on gbdev.io](https://gbdev.io/pandocs/The_Cartridge_Header.html) and [the GBDev wiki](https://gbdev.gg8.se/wiki/articles/The_Cartridge_Header).

gmestanley:
Added a hotfix for gb.hexpat as well, fixing not only the syntax error but also the CGB bitfield not being able to be formatted. I really need to not submit patterns without testing them...

WerWolv:
Thank you!

PR #177, opened 2023-10-21
zip pattern: Improved fallback method for finding eocd

flawwan:
Hi :)

I found a bug when I ran the zip pattern on some random zip file I had. In this case it was the zip of the Binwalk 2.3.4 source code. You can verify by download it from github (releases and then download (Source code(zip)). Open with imHex and select zip pattern. It will fail. I created my own zip test data to cover this bug.

The bug is when multiple signatures for the eocd are located in the last 64kb. The fallback will pick the first one, which does not always work when there might be multiple potential eocd signatures within that range.

This pull request fixes that bug and should make the fallback method more reliable. This is done by looping through all eocd signatures within the last 64kb. Invalid "eocd" signatures will be discarded and the correct eocd will be returned.


WerWolv:
Thank you!

PR #178, opened 2023-10-23
fix payload size when ip header or tcp header has options in ip.hexpat

leenzhu:
The orginal code was considering the IP header and TCP header has fixed lenght of 40 bytes. Actually it should be calculated by related header fileds in order to support any valid TCP packets.


WerWolv:
Thank you!

PR #182, opened 2023-10-24
zip pattern: added parsing of extra field

flawwan:
Hi!

I added parsing for the extra field in the zip pattern. The following tags are now supported:

- 0x5455 (ExtendedTimestamp)
- 0x000A (NTFS)
- 0x7875 (NewUnix)
- 0x5855 (InfoZipUnix)


WerWolv:
Thank you!

PR #190, opened 2023-11-11
Added pattern file for Apple’s binary property list format (bplist)

flawwan:
This pull request adds pattern support for Apple’s binary property list format (bplist).


WerWolv:
Thanks a lot!

PR #193, opened 2023-11-13
MP4: make handler_type into a char[4], which reflects how it is used

klirktag:
handler_type is actually a 4 character string, so it better be represented as a char[4] (instead of u32)
example values are "soun" and "vide", here is a list of possible values

https://cconcolato.github.io/mp4ra/handler.html

Screenshot of handler_type="soun" (when changed to char[4])
Screenshot from 2023-11-13 17-45-03


WerWolv:
Thank you!

PR #194, opened 2023-11-13
patterns/hinf_bitmap: Added pattern for parsing Halo Infinite bitmap files

Surasia:
- Parses the header of all Halo Infinite files which includes related "dependencies", datablocks, structure references
- Read bitmap sequence data, including bitmap index, complete index, sprite info
- Find the width/height and DDS Format of the raw texture, which can then be parsed into usable DDS textures
- Locates raw DDS block info


WerWolv:
Thanks a lot! 
Just to be safe, is the test file you provided directly from the Game? If so, could you maybe strip it down so it doesn't contain any original image data anymore? Just to avoid DMCA takedowns

Surasia:
done! Now only contains the header information.

WerWolv:
Perfect thank you!

PR #195, opened 2023-11-13
MP4: Support mp4a and avc1 codec boxes

klirktag:
Support for avc1(h264) and mp4a(aac) codecs

Can extract things like width and height of the video, and sample_rate of the audio

Example of Avc1 info (width and height)
Screenshot from 2023-11-14 00-20-10

Example of mp4a (sample_rate)
Screenshot from 2023-11-14 00-21-07

Also the (sample)TimeToSample "stts"box is implemented
Screenshot from 2023-11-14 15-38-07

### DIFF
--- a/patterns/mp4.hexpat
+++ b/patterns/mp4.hexpat
@@ -16,7 +16,17 @@ fn to_string(auto var) {
 
 fn format_string(auto string) {
     return string.value;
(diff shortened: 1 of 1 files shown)


WerWolv:
Thank you!

PR #198, opened 2023-11-15
fix:  [ImHex Bug] std::string::replace("aa", "aa", "") returns "a" #1408

paxcut:
The previous fixes for this function didn't consider the case when the resulting string has a contribution from the original string that occurs after all the substitutions have been made and that is longer than the replacement string. This PR resolves that issue and also addresses the case where one or more of the inputs is empty. Please see discussion on [ImHex issue #1408](https://github.com/WerWolv/ImHex/issues/1408) for details.


WerWolv:
Thank you!

PR #203, opened 2023-11-20
patterns/hinf_luas: Added pattern for reading HavokScript 5.1 Bytecode from Halo Infinite

Surasia:
* Parses Header and Structures of the parent tag
* Reads Bytecode types and arguments
* Reads Constants and Debug Info
* Reads Referenced Tags in the File

(in comparison to the other Halo Infinite pattern I submitted, this test file only contains raw bytecode and debug info which should not contain any copyright material, however I have a nulled out version as well.)


WerWolv:
Thank you!

PR #204, opened 2023-11-21
Bringing over the massively improved version of pe.hexpat

gmestanley:
Hi! Here is the new version of pe.hexpat I concocted.

Differences are:
- Different programming choices taken in favor of performance improvements (e.g. using `!variable` instead of `variable == 0`
- Rich header visualization made better (products now have their actual IDs listed instead of their masked ones)
- Clearer variable names, including ones where the spec name isn't clear enough
- Added spec names from DOSHeader to SectionFlags (not to enums because they don't support it)
- Inaccuracies fixed ("None" isn't listed as a DataAlignment value in SectionFlags in Microsoft's own documentation, it was just a workaround in case it's not set)
- Types like guid and time used where applicable 
- RichHeader made so that it doesn't need to be an array to only be displayed on Pattern Data if it exists
- Minor data structures such as RSDS added
- Data directories now make $ go to the end of them to avoid inaccurate size displayed on Pattern Data window
- Functions like finddosmessage and formatSectionName made simpler
- How exactly the Bitmap struct's image data works is taking a while to be understood, so a u8 array is filling lineData's place


WerWolv:
Thank you!

PR #205, opened 2023-11-26
Create gltf.hexpat

utkumaden:
Added glTF binary file pattern.

This is my first time writing a ImHex pattern so any feedback is welcome. Although glTF files are mostly JSON, the binary format allows you to encode the binary buffers directly in the file which is useful for performance and load times. I quickly wrote this file to confirm a file I had matched the official quick start guide. Humble, but useful.


patterns/gltf.hexpat
```diff
@@ -0,0 +1,66 @@
...
+ * @brief The glTF magic section.
+ */
+struct gltf_magic_t {
+    char magic[4];  /**< The magic value. Must be "glTF" */
+    u32 version;    /**< The version. Must be 2 for glTF 2.0. */
+    u32 length;     /**< Length of the file in bytes, including magic section. */
+};
+
+/**
+ * @brief Enumeartion of well defined glTF chunks.
```
mgerhardy, line 38: typo

patterns/gltf.hexpat
```diff
@@ -0,0 +1,66 @@
...
+    JSON = 1313821514 /* "JSON"  */,  /**< JSON chunk. */
+    BIN  = 5130562    /* "BIN\0" */,  /**< Binary data chunk. Could be an image or model data. */
+};
+
+/**
+ * @brief A glTF chunk.
+ */
+struct gltf_chunk_t {
+    u32 length;             /**< Length of this chunk. */
+    gltf_chunk_type_t type [[format("gltf_format")]]; /**< Type of the chunk. JSON or BIN expected. */
```
mgerhardy, line 50: wouldn't it be better to enforce the endianness here because we are checking against the u32 representation.

WerWolv:
Thank you! I'll leave this open for you to respond to the review

utkumaden:
I had a busy week at work and forgot about this. My bad, let's fix this up.

patterns/gltf.hexpat
utkumaden: Possibly, we know glTF files are always little endian. I'm not familiar with endianness in imhex patterns, so a pointer would be much appreciated.

WerWolv: If the format always uses little endian, then there's nothing to worry about

WerWolv:
Merged, thanks a lot!

PR #225, opened 2024-03-01
fix:"unknown function fm" error was shown in pattern data.

paxcut:


Hexcede:
Correct me if I am wrong, but this seems to be a bug with the pattern language. When importing the library it should be able to work within that library's scope.

paxcut:
I think that anything defined inside a namespace needs to be fully qualified to be able to use it, even if that is  from within the namespace itself. The new import statement allows you to alias namespaces but that's very new addition and as far as I know you always needed to include the namespace resolution or else get an error. For example, this simple test fails if you omit the 'test::'
```cpp
namespace test {
    fn test(auto a) {
        return a+1;
    };
    
    struct A {
        u32 a = test::test(2);
    };
}

test::A a@0;
```

WerWolv:
Thank you!

paxcut:
Actually @Hexcede was right. I checked if structs could be used without qualification inside their namespace and the could which seems inconsistent so i submitted an issue to the pattern language repository (number 85).

PR #233, opened 2024-03-21
Replace get_attribute_value with get_attribute_argument

ZwipZwapZapony:
In early 2023, `builtin::std::core::get_attribute_value(pattern, attribute)` was changed to `builtin::std::core::get_attribute_argument(pattern, attribute, index)`, and a third parameter was added to that function.
"*includes/std/core.pat*" was never updated for either of those things, causing compile errors when using `std::core::get_attribute_value` in a pattern.

This pull request updates that file to add `std::core::get_attribute_argument(pattern, attribute, [index])` (with `index` defaulting to 0), and makes `std::core::get_attribute_value` show you an error telling you to use `std::core::get_attribute_argument` instead. (I would've made `std::core::get_attribute_value` use an index of 0, but given that it's been broken for over a year, I don't think that a lot of current patterns use it.)

(This pull request also updates the library documentation comments to have square brackets around optional arguments and list their default values, to correspond with https://github.com/WerWolv/Documentation/pull/12.)

\
Note that this is untested; ImHex tells me `runtime error: Cannot cast value to type 'pattern'` while evaluating a pattern where I try to use pattern-related functions. Therefore, I don't know whether `index` should default to 0 or to 1, so I just assumed 0.

### DIFF
--- a/includes/std/core.pat
+++ b/includes/std/core.pat
@@ -38,12 +38,20 @@ namespace auto std::core {
     };
 
     /**
-        Returns the first parameter of the attribute of a pattern if it has one
+        Returns the nth parameter of the attribute of a pattern if it has one
         @param pattern The pattern to check
         @param attribute The attribute's name to query
+        @param [index] The parameter index of the attribute to return. Defaults to 0
+    */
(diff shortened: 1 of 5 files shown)


WerWolv:
Thanks a lot!

Issue #234, opened 2024-03-25
obscure pattern welcome?


ccoenen:
I own a device that's pretty old and no longer manufactured. It has a pretty simple file format that I created a pattern for to learn the pattern language. Should I contribute it? I doubt that it has *practical* use, but it is a very simple format and might serve as an example. The whole file is an array of the same `struct` and it also has a few bitfields.

WerWolv:
Hey!
Generally everything is welcome as long as the pattern _might_ have any use for someone else.

Obscure formats that are only useful in very specific cases are therefor completely fine, even if they might only help very few people.
Formats that are only used internally in e.g the company you work for are the exception here since these most likely shouldn't be made public anyways and also probably would only help you and maybe your work colleagues.

For your case, I think definitely go for it :)

PR #238, opened 2024-04-07
Adding the hexpat for Wintec TES files

ccoenen:
With this pull request, I'm adding support for a very simple (if somewhat obscure) file format for a specific GPS Tracker from the early 2000s. The device itself is no longer being sold.

The pattern itself makes use of a lot of the functionality on top of this simple file format. Maybe it is useful as an example for other pattern files. Specifically, it adds colors, comments and formatters for most fields.

tooltip for latitude in degrees

coordinate preview window

Also referencing discussion #234


WerWolv:
Thank you!

Issue #239, opened 2024-04-14
[Feature] Obvious patterns for overridden JPEGs.


vodkar:
Hello!
## Feature description
I'd like a feature if JPEG Pattern Parser will show correctly shows overridden JPEGs.
## Example
Example of such image:
![orig_schematic_fixed_53c1dec](https://github.com/WerWolv/ImHex-Patterns/assets/28678409/585ad92c-4575-4255-a30b-f3d2b6aafdd8)
This is how it processed now: 
![Screenshot 2024-04-15 at 00 56 31](https://github.com/WerWolv/ImHex-Patterns/assets/28678409/82ac44fb-0749-43ca-ad8e-164a982dc8a0)
## Expected behavior
And this is how it expected to be processed (screenshot from 010 Editor):
![image](https://github.com/WerWolv/ImHex-Patterns/assets/28678409/165d6f2a-9296-4d17-99a6-2a192a76c1a6)

PR #240, opened 2024-04-14
Support EOF in the middle of JPEG

vodkar:
- Support for EOF if EOF in the JPEG met not in the end.

## Before

image

## After

image

## JPEG to reproduce

Schematic 53c1dec

More information in #239


WerWolv:
Hey! Thanks a lot for the PR

There seems to be an issue with your changes that makes the CI fail to run with the jpeg test file we're using. Could you check what happened there?

vodkar:
I'm not sure, but I think there is a bug in `find_sequence` from pattern language.

I am using this [image](tests/patterns/test_data/jpeg.hexpat.jpeg) for tests and `std::print` for debug:
```
struct SOS {
    u8 numComponents;
    SOSComponent components[numComponents];
    u8 startSpectralSelection;
    u8 endSpectral;
    u8 apprBitPos;
    
    std::print("{}", std::mem::find_sequence(0, 0x8F, 0xD0, 0xFF, 0x00, 0xFF));
    u8 image_data[std::mem::size() - $ - 2] [[sealed]];
};
```
And I got this result, and it's correct:
image

But when I change `std::print("{}", std::mem::find_sequence(0, 0x8F, 0xD0, 0xFF, 0x00, 0xFF));` to `std::print("{}", std::mem::find_sequence(0, 0x8F, 0xD0, 0xFF, 0x00, 0xFF, 0xD9));` I got a wrong result:
image
Looks like `find_sequence` is not checking the most last byte.
I ask ChatGPT about [this code](https://github.com/WerWolv/PatternLanguage/blob/83b2686e71327c73bc8a12afa5e09760351152b3/lib/source/pl/lib/std/mem.cpp#L15), and this is what I got:


image


It's hard for me to confirm this, but can I kindly ask you to check this bug?

paxcut:
It wasn't that hard to verify the error. After getting -1 when the last byte of the file was included in the search I resized the input file by adding an extra byte at the end and then the  search including 0xD9 was correct again.

 It looks like the unit test fails because it uses the result of the find  sequence function to calculate an array size. Because of the bug you found, the find sequence function is returning -1 which is not a valid array size. Although all jpegs must include this sequence it is a good idea to check the result of the find sequence function for failure to find it (eg file was corrupted) so you can print a better error message (like "missing eof sequence" or something)  instead of the "-1 is not a valid array size" that it prints currently.

vodkar:
Ready to check

vodkar:
Ping

WerWolv:
Thank you!

PR #243, opened 2024-05-05
patterns: add Digital Terrain Elevation Data (DTED) pattern file

ckoehler:
Tested on DTED0 and DTED2, will likely work with DTED1 as well. Adding a DTED0 file for testing.


WerWolv:
Thank you!

PR #246, opened 2024-05-12
Correctly call builtin flush

mandar1jn:
The std::file::flush function called the builtin remove function instead of the builtin flush function, meaning that any call to flush would actually delete the file


WerWolv:
My god how did I not catch that. Thank you!

PR #247, opened 2024-05-12
Add add_virtual_file to hex/core.pat

mandar1jn:
This function is implemented [here](https://github.com/WerWolv/ImHex/blob/master/plugins/builtin/source/content/pl_builtin_functions.cpp#L32-L45), yet is never directly exposed using the hex/core.pat pattern file.


WerWolv:
Thanks a lot!

PR #248, opened 2024-05-13
mp4: Use FixedPoint32 instead of u32 for width and height

Shimogawa:
According to ISO/IEC 14496-12, the width and height fields should be fixed point 16.16.

* Use FixedPoint32 instead of u32 for width and height
* Add formatting for displaying fixed point 16.16 and 8.8 numbers


WerWolv:
Thank you!

PR #255, opened 2024-05-22
Updated the custom themes to the new text editor color palettes 

paxcut:
Even though the new palettes were designed for full syntax highlighting, the existing themes should look just like they did before. Full syntax highlights will need the use of new themes that are added in other repositories.


paxcut:
This pull request should  be easy to review but hasn't seen any comments for 3 months. I don't mean to be pushy and this is certainly not critical stuff but as the new policy for dealing with issues that are old has been put in effect and seems to be showing signs of improvement,  I think it isn't a bad idea to implement a similar approach for PRs that can be considered to be stale. Asking to be reviewed every three months should be reasonable I think but if in disagreement by all means suggest a better time interval.

PR #257, opened 2024-05-26
patterns/pe added .didata section support

qknight:
This code works and can be used with binaries using PE (exe files) and a delayed dll loader section called .didata:

grafik


WerWolv:
Thanks a lot!

qknight:
@WerWolv would you be ok if i also contribute the .fixPath section parser even if it is not a PE standard... yet? See https://lastlog.de/blog/libnix_fixPath.html

PR #259, opened 2024-06-09
Fix bitfield PixelFormatFlags

sisco0:
The padding names should not repeat
Comments should be added to each field for clarification
Padding should be corrected for luminance.

# Pattern

[Information about your pattern]

## Checklist
- [ ] A pattern for this format doesn't exist yet (or this PR improves the existing one)
- [ ] The new pattern has been added to the relevant table in the Readme
- [ ] The new pattern has a description pragma (`#pragma description My pattern Description here`)
- [ ] The pattern was associated with all relevant MIME types (using `#pragma MIME mime-type` in the source code)

### DIFF
--- a/patterns/dds.hexpat
+++ b/patterns/dds.hexpat
@@ -167,16 +167,16 @@ bitfield Caps2Flags {
 };
 
 bitfield PixelFormatFlags {
-  alphaPixels : 1;
-  alpha       : 1;
-  fourCC      : 1;
-  padding     : 3;
-  rgb         : 1; // 0x40
-  padding     : 2;
-  yuv         : 1; // 0x200
-  padding     : 3;
-  luminance   : 1; // 0x20000
-  padding     : 17;
+  alphaPixels : 1;  // DDPF_ALPHAPIXELS: 0x00000001
+  alpha       : 1;  // DDPF_ALPHA: 0x00000002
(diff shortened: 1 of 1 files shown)


WerWolv:
Thanks a lot!
Is it important that the values of the padding fields are shown? `padding` is a special keyword otherwise that simply skips those bits without them being shown in the pattern data view.

sisco0:
Thank you for your kind review @WerWolv .
I have now re-worked the current Pull Request with proper values after consulting [GIMP](https://gitlab.gnome.org/GNOME/gimp/-/blame/master/plug-ins/file-dds/dds.h#L145) repository.

sisco0:
I would re-factor the current Pull Request to contain a test, so ideally I could learn more from the repository structure and could provide evidence.

WerWolv:
There's already a test file in the repo that the current pattern is being run on to make sure it runs correctly. But you can add some `std::assert()` calls to the pattern file to verify values if there's any good candidates

sisco0:
I attach graphic evidence of the file under `test` folder with positive results compared to current [GIMP](https://gitlab.gnome.org/GNOME/gimp/-/blame/master/plug-ins/file-dds/dds.h#L145) implementation in this comment:
Evidence_Pixelformatflags

PR #263, opened 2024-06-29
Add ttf/otf format

Shimogawa:
Specs:
- https://learn.microsoft.com/en-us/typography/opentype/spec/
- https://developer.apple.com/fonts/TrueType-Reference-Manual/RM06/Chap6.html

The test file added is [Roboto](https://github.com/googlefonts/roboto)-Regular.ttf, opensource with [Apache 2.0](https://github.com/googlefonts/roboto/blob/main/LICENSE).


WerWolv:
Fantastic work, thank you very much!

PR #266, opened 2024-07-03
Add Password Safe Version 3 pattern

mockbutler:
Checks for valid identifier tag and end of data marker.
Password for sample file is: "password".


WerWolv:
Thanks a lot!

mockbutler:
@WerWolv Yeah sorry it was a bit of mess. Tried compiling the pattern test on Mac, but the xcode version of clang (Apple clang version 15.0.0 (clang-1500.3.9.4)) fails on the deprecated functions:
```
pattern_language-src/lib/include/pl/patterns/pattern_wide_character.hpp:34:32: error: 'wstring_convert<std::codecvt_utf8_utf16<char16_t>, char16_t>' is deprecated [-Werror,-Wdeprecated-declarations]
```
And it was taking me some time to find where to add `-Wno-deprecated-declarations` to see if I could fix the failed test.

WerWolv:
No worries! I had to look into the testing code too actually to see why it was failing. I added a better error message there now.

Apple clang is generally unsupported because it's so majorly out of date constantly. I can however replace that `wstring_convert` nowadays though I think

PR #267, opened 2024-07-06
patterns/hinf_*: Add Halo Infinite tag and module file patterns

Surasia:
Added both the main archive format and serialized data format for Halo Infinite. Both patterns doesn't require updating with game versions, as opposed to the hinf_bitmap pattern which I have removed. I've also updated hinf_luas to be more accurate, which should now work with any file.

Test files are stripped version of files extracted from the game.


WerWolv:
Thanks a lot!

mandar1jn:
A bit late but aren’t the md links wrong? They refer to the deleted file

Surasia:
you're right, forgot to change them :( will update soon in another PR! thanks for letting me know

PR #274, opened 2024-07-23
Add PKM pattern + add BCC checksum verification to ntag pattern

applecuckoo:
Hi! First time here, but probably won't be the last :)
The PKM pattern was something that I wrote just to learn the ropes, and the NTAG checksum was quite simple so I went about implementing that.


WerWolv:
Thank a lot! :)

PR #276, opened 2024-07-24
patterns: FBX - fixed the fbx hexpat

Hikodroid:
I split the fix and some improvements into two separate commits. Both should be functionally the same and I would have expected both to be fully working but the first commit does not work and the second one does. I'm not sure why my weird workaround actually works but it should be in a useful state now.

### DIFF
--- a/patterns/fbx.hexpat
+++ b/patterns/fbx.hexpat
@@ -4,7 +4,10 @@
 
 #pragma endian little
 
-import hex.dec;
+#ifdef __IMHEX__
+    import hex.dec;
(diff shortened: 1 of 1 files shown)


WerWolv:
It's not working in the CI because it's an ImHex extension, it's not available outside of it.
This is actually the first pattern that makes use of functions like this that's on the repo. I don't think it should be removed but I need to think of a way to make it still work here too. Maybe with an #ifdef or something. I'll take care of that though, nothing for you to worry about

PR #277, opened 2024-07-25
patterns: FBX - simplified the fbx hexpat

Hikodroid:
The switching between A and B type NodeRecords turned out to be unnecessary.


WerWolv:
Thanks a lot!

PR #281, opened 2024-07-29
Added pattern for Flipper Zero settings

janwiesemann:
Added pattern to support parsing of Flipper Zero settings.


WerWolv (commented):
Added readme entry

WerWolv:
Thank you!

PR #285, opened 2024-08-03
patterns: Added Lua 5.3 bytecode pattern

qux-bbb:


WerWolv:
Thanks!
Would it make sense maybe to combine all these versions into a single pattern? The version information seems to be in the binary file and the format of all looks reasonably similar

qux-bbb:
There are some differences between these version, so it's difficult to combine them into a single pattern.

WerWolv:
Do you possibly also have a Lua 5.4 test file? That's one of the few test files we're missing still

qux-bbb:
> Do you possibly also have a Lua 5.4 test file? That's one of the few test files we're missing still

I'll upload a Lua 5.4 test file.

paxcut:
i've noticed that in 5.3 and 5.4 the values of type that decide if value is int or floating point are reversed. Did they decide that swapping them provided some sort of advantage?

Lua 5.3
```cpp
    } else if (type == 3) {  // LUA_TNUMFLT
        double data;
    } else if (type == 0x13) {  // LUA_TNUMINT
        u64 data;
```
Lua 5.4
```cpp
    if (type == 3) {
        u64 data;
    } else if (type == 0x13) {
        double data;
```

qux-bbb:
Yes.  

Lua 5.3.6  
```c
/* Variant tags for numbers */
#define LUA_TNUMFLT	(LUA_TNUMBER | (0 << 4))  /* float numbers */
#define LUA_TNUMINT	(LUA_TNUMBER | (1 << 4))  /* integer numbers */

    case LUA_TNUMFLT:
      DumpNumber(fltvalue(o), D);
      break;
    case LUA_TNUMINT:
      DumpInteger(ivalue(o), D);
      break;
```

Lua 5.4.6  
```c
/* add variant bits to a type */
#define makevariant(t,v)	((t) | ((v) << 4))

#define LUA_TNUMBER		3

/* Variant tags for numbers */
#define LUA_VNUMINT	makevariant(LUA_TNUMBER, 0)  /* integer numbers */
#define LUA_VNUMFLT	makevariant(LUA_TNUMBER, 1)  /* float numbers */

      case LUA_VNUMFLT:
        dumpNumber(D, fltvalue(o));
        break;
      case LUA_VNUMINT:
        dumpInteger(D, ivalue(o));
        break;
```

https://www.lua.org/ftp/lua-5.3.6.tar.gz  
https://www.lua.org/ftp/lua-5.4.6.tar.gz

paxcut:
I looked into combining all the lua versions into 1 pattern and I can probably do that if there is any interest.

PR #287, opened 2024-08-07
patterns: Added RGBDS object file format

ISSOtm:
Hello! Thank you for this very fun hex {viewer, editor, ...}

As part of trying it out, I wrote a pattern for a file format I use *very* often now. Hope you'll like it :)


ISSOtm:
(Note: I don't know why, but all of the `NullString`s fail to display, with `runtime error: No variable named 'data' found.`.)

WerWolv:
Thanks a lot!

> (Note: I don't know why, but all of the NullStrings fail to display, with runtime error: No variable named 'data' found..)
That's my fault, easy enough to fix though

PR #288, opened 2024-08-09
patterns/uf2: Add new RP2xxx-related family IDs

applecuckoo:
I think I probably didn't expect a handful of new family IDs to be added to UF2 a mere two weeks after my initial patch for the UF2 pattern... maybe the more surprising thing is that for the first time, there are multiple family IDs for the exact same chip. Here are those new family IDs, fresh from the file spec.


WerWolv:
Thanks a lot!

PR #289, opened 2024-08-10
Update core.pat - fix set_pattern_comment

BobSmun:
Probably a copy-paste issue - std::core::set_pattern_comment has a parameter `comment`, but was using (and the comment was documenting) `name`


WerWolv:
Thank you!

PR #291, opened 2024-08-15
Fix msgpack bugs

SpaceManiac:
* 16-bit and 32-bit extension lengths were not marked big-endian.
* Negative fixint values were being mangled.


WerWolv:
Thank you!

PR #293, opened 2024-08-19
patterns: Added ext4 filesystem pattern

endes0:


WerWolv:
Very cool, thanks a lot!

PR #294, opened 2024-08-20
quick readme fix + new WebP and VGM patterns

### LINKED ISSUE #152: Protobuf pattern does not support field numbers > 15
As per [Protobuf Documentation](https://protobuf.dev/programming-guides/encoding/#structure), "Tags" (Keys) of Records/Keys are encoded as VarInts / LEB128.

However the current Protobuf pattern assumes Keys are always 1 byte:
https://github.com/WerWolv/ImHex-Patterns/blob/6baae925532a8e3fb557fd07273063c306e5bc5b/patterns/protobuf.hexpat#L34

This assumption will not work anymore for larger (>15) field numbers, where the varint will be 2 bytes long:

```java
CodedOutputStream stream = ...;
stream.writeUInt(16, 0);
```

The output binary will be:

```
80 01 00
```

The value byte here is expected to be 0. The current pattern will incorrectly interpret the second byte as the value 1.

### LINKED ISSUE #180: request: VGM pattern
Hi

I would like to request a VGM pattern (video game music)
You can see the header here

https://vgmrips.net/wiki/VGM_Specification

I'm a total beginner and can't make this myself :(
I hope somebody can create this pattern (for latest VGM version)

### LINKED ISSUE #261: Protobuff definition is not correct
https://github.com/WerWolv/ImHex-Patterns/blob/3416d30f2b8fac4f12bf73aa1f5d5102a07b7f83/patterns/protobuf.hexpat#L34C5-L37C4

The [specification](https://protobuf.dev/programming-guides/encoding/) for protobuff states that this record is Varint Encoded. 

So it means that it's not guaranteed to have 8 bits. If the field index has its last significant bit on 1 then another byte has to be read.

I've tested this in real life payloads and I've confirmed that it's failing.

applecuckoo:
this PR is getting bigger and bigger as it is left unmerged, lol

Also, the VGM pattern should fix #180. There's no VGM test file as I'm not sure if there are any public domain/GPL v2 examples out there.

The Protobuf commits should fix #261 and fix #152.


C3pa (approved):

applecuckoo:
@WerWolv hey! Just a quick reminder ping to merge this and all the other PRs sitting around. Also, a couple of test file notes from me - I've made it tickable for ease of tracking:

- [ ] I haven't put in a VGM sample due to potential copyright pitholes and stuff. The stuff at [Battle of the Bits](https://battleofthebits.com/) seems to be mostly CC BY-NC-SA, so I can't use that, and the [Furnace tracker demos](https://github.com/tildearrow/furnace/tree/master/demos) are all rights reserved, so that's out of the picture as well.
- [ ] The protobuf test file is just full of zeros and doesn't really appear to be a 'real' protobuf wire, might be worth investigating

WerWolv:
Thank you! Sorry for the huge delay...

PR #295, opened 2024-08-21
Adding checks to ends_with and starts_with to prevent failure

MartorSkull:
Fixing cases where part is longer than string, this cases should both return false instead of failing.

E.g.:
```rust
ends_with("asdf", "0123456789");
```
fails with
```
E: runtime error: Character index 4294967290 out of range of string 'asdf' with length 4.
```
but it should simply return `false`.

This should be merged together with a fix to the builtin substr function to throw an error if the desired substring does not fit inside the string.


WerWolv:
Thank you!

PR #298, opened 2024-08-27
Fixed a bug in the Lua 5.1 bytecode pattern

Guest257351:
Lua 5.1 bytecode upvalue fields were incorrectly defined as a `Vector<u32>` instead of `Vector<LuaString>`.

### DIFF
--- a/patterns/lua51.hexpat
+++ b/patterns/lua51.hexpat
@@ -72,7 +72,7 @@ struct LocalVar {
 struct LuaDebugInfo {
     Vector<u32> lineInfo;
     Vector<LocalVar> localVar;
-    Vector<u32> upvalues;
+    Vector<LuaString> upvalues;
 };
 
 struct LuaConstants{


paxcut:
I know very little about Lua byte code, but having separate patterns for each version of Lua makes dealing with problems like this a pain. Instead of checking every other Lua to see if the error exists there I unified all the Lua versions into one. The resulting pattern has this definition for the upValues
```cpp
struct UpValues {
    if (minorVersion == 1)
        Vector<u32> upvalues;
    else
        Vector<LuaString> upvalues;
}[[inline]];
```
Not only shows a potential error but also suggests what the fix would be and all of this without having to visit each and every version there is. I am attaching the unified version (change the extension that github cant deal with to .hexpat) I wrote which I tested successfully with all the tests for lua versions ImHex has, but probably needs more rigorous tests given that this error was not found.
[lua.txt](https://github.com/user-attachments/files/16760083/lua.txt)

Guest257351:
> I know very little about Lua byte code, but having separate patterns for each version of Lua makes dealing with problems like this a pain. Instead of checking every other Lua to see if the error exists there I unified all the Lua versions into one. The resulting pattern has this definition for the upValues
> 
> ```c++
> struct UpValues {
>     if (minorVersion == 1)
>         Vector<u32> upvalues;
>     else
>         Vector<LuaString> upvalues;
> }[[inline]];
> ```
> 
> Not only shows a potential error but also suggests what the fix would be and all of this without having to visit each and every version there is. I am attaching the unified version (change the extension that github cant deal with to .hexpat) I wrote which I tested successfully with all the tests for lua versions ImHex has, but probably needs more rigorous tests given that this error was not found. [lua.txt](https://github.com/user-attachments/files/16760083/lua.txt)

As far as I'm aware lua does not use u32s or integers of any kind for it's upvalue vectors. I don't know of any lua version that does. The upvalue vector is only there for debug information.
As for having multiple version being a pain, I might look into how lua creates bytecode for versions 5.1 onward to see if they can easily be put into a single pattern.

Guest257351:
Just added a commit for a separate similar bug. I forgot to include this in the initial pull request. It's also worth noting that the number of bytes in the string size is not static. By default lua uses 32-bit integers and just about every implementation of it also does since there's no good reason to change it. The number of bytes is controlled by `size_of_size_t` in `LuaBinaryHeader` so it might be worth changing the pattern to support this value being anything other than `4`.

paxcut:
I put a link to the unified version of the lua pattern. Did you look at it?

paxcut:
I just checked the sample Lua 5.1 sample that is used in unit tests and the string is preceded by a 64 bit integer holding its size. The change that you propose turns a successful test into a failing one of reading past the end of the file. Make sure to test the changes on actual files and if you think the ones we use are invalid provided valid ones.

Guest257351:
> I just checked the sample Lua 5.1 sample that is used in unit tests and the string is preceded by a 64 bit integer holding its size. The change that you propose turns a successful test into a failing one of reading past the end of the file. Make sure to test the changes on actual files and if you think the ones we use are invalid provided valid ones.
> 
> image

I did test the patterns before making these changes. After looking into how lua defines size_t, the test samples were created from a lua 64 bit compiler, where as I was using a 32 bit compiler. The unified pattern you sent does seem to account for size_t's variable size, so I guess this was an issue specific to the lua 5.1 pattern. I'll update my pull request to account for size_t's variable size.

paxcut:
>  The unified pattern you sent does seem to account for size_t's variable size, so I guess this was an issue specific to the lua 5.1 pattern. I'll update my pull request to account for size_t's variable size.

I don't follow. The unified pattern is exactly equivalent to each of the separate versions, or at least , it aims to be. If one of the versions has a problem, then the unified version will have the exact same problem. For example, for string size type it defines:
```cpp
struct LuaStringSize {
   if (minorVersion == 4)
        Size size;
    else if (minorVersion == 3)
        u8 size;
    else
        u64 size;
}[[sealed, format("impl::format_StringSize"), transform("impl::transform_StringSize")]];
```
So both versions 1 and 2 use u64 like the 5.1 and 5.2 versions do. Im looking into getting better samples so we can test things better.

Guest257351:
> > The unified pattern you sent does seem to account for size_t's variable size, so I guess this was an issue specific to the lua 5.1 pattern. I'll update my pull request to account for size_t's variable size.
> 
> I don't follow. The unified pattern is exactly equivalent to each of the separate versions, or at least , it aims to be. If one of the versions has a problem, then the unified version will have the exact same problem. For example, for string size type it defines:
> 
> ```c++
> struct LuaStringSize {
>    if (minorVersion == 4)
>         Size size;
>     else if (minorVersion == 3)
>         u8 size;
>     else
>         u64 size;
> }[[sealed, format("impl::format_StringSize"), transform("impl::transform_StringSize")]];
> ```
> 
> So both versions 1 and 2 use u64 like the 5.1 and 5.2 versions do. Im looking into getting better samples so we can test things better.

Sorry I've made a mistake again.😅The unified version does not account for 32-bit vs 64-bit compilers, I probably should have read through it more thoroughly.

Guest257351:
Also I'm not sure how I'd easily make either the unified version or the 5.1 version account for this. Ideally `LuaString` should use a `u32` if `LuaBinaryHeader.size_of_size_t` is 4, and a `u32` if it's 8. However I'm not experienced enough with the pattern language to implement this.

paxcut:
something like this should work:
create a global 
```cpp
u8 global_size_of_size_t=4;
```
we assign a default just in case. Then inside header, right after size_of_size_t is created
```cpp
global_size_of_size_t = size_of_size_t;
```
finally on StringSize
```cpp
struct LuaStringSize {
   if (minorVersion == 4)
        Size size;
    else if (minorVersion == 3)
        u8 size;
    else {
        if (global_size_of_size_t == 4 )
            u32 size;
        else
            u64 size;
    }
}[[sealed, format("impl::format_StringSize"), transform("impl::transform_StringSize")]];
```
I am showing the code in unified pattern but it should be the same for 5.1. What about the other versions?

Guest257351:
> something like this should work: create a global
> 
> ```c++
> u8 global_size_of_size_t=4;
> ```
> 
> we assign a default just in case. Then inside header, right after size_of_size_t is created
> 
> ```c++
> global_size_of_size_t = size_of_size_t;
> ```
> 
> finally on StringSize
> 
> ```c++
> struct LuaStringSize {
>    if (minorVersion == 4)
>         Size size;
>     else if (minorVersion == 3)
>         u8 size;
>     else {
>         if (global_size_of_size_t == 4 )
>             u32 size;
>         else
>             u64 size;
>     }
> }[[sealed, format("impl::format_StringSize"), transform("impl::transform_StringSize")]];
> ```
> 
> I am showing the code in unified pattern but it should be the same for 5.1. What about the other versions?

That could work, I've managed to implement it differently. You can review that and see which works better.

paxcut:
Why are you only fixing lua 5.1? Are the other Luas also not possibly erroneous? Considering that we have to deal with 4 versions, there should be an attempt to keep them similar to each other to help debug issues that deal with more than just one version, so breaking structures that are kept together in the other versions doesnt seem like a good idea.

The main issue is why don't you think it is a good idea to get rid of the version files and use one instead? That avoids code duplication, reduces code management, makes it more clear what is the same and what isn't across versions and helps showcasing the pattern language ability to deal with special cases  among other things. What are the downsides?

paxcut:
A better solution than both
```cpp
u8 stringSize  =  std::mem::read_unsigned((minorVersion>2) ? 13 : 8,1);
```
eliminates the need to copy from the pattern to a global or the need to break the main placed variable into pieces.

Guest257351:
> Why are you only fixing lua 5.1? Are the other Luas also not possibly erroneous? Considering that we have to deal with 4 versions, there should be an attempt to keep them similar to each other to help debug issues that deal with more than just one version, so breaking structures that are kept together in the other versions doesnt seem like a good idea.
> 
> The main issue is why don't you think it is a good idea to get rid of the version files and use one instead? That avoids code duplication, reduces code management, makes it more clear what is the same and what isn't across versions and helps showcasing the pattern language ability to deal with special cases among other things. What are the downsides?

We should combine the versions, I didn't do that because it's not what the original issue was trying to fix. But yeah absolutely we should.

PR #299, opened 2024-09-14
fix off by one error in coff pattern

vlenhart:
this now behaves like the pe pattern
see specification: https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#section-flags


WerWolv:
Thank you!

PR #300, opened 2024-09-14
use appropriate pointer for first section placement

vlenhart:
while the previous way seemed to work it should be more robust to use the ptrRawData to place the start of sections

### DIFF
--- a/patterns/pe.hexpat
+++ b/patterns/pe.hexpat
@@ -1106,7 +1106,7 @@ struct Section {
 	currentSectionIndex += 1;													// Make the current section index the next section's index
 } [[name(sectionsTable[currentSectionIndex-1].name)]];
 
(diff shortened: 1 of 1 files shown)


paxcut:
Could you explain a bit the basis for the argument on robustness? In other words, why is the new method preferable over the previous one? Are there cases where the old method would fail to obtain the starting address of sections that the new method would fix? I have an interest on the PE format  so if there is something like that I would appreciate learning about it.

vlenhart:
While I think that most PE files out there probably have the section data placed immediately after the section headers, the specification does not require this. The specification only states that the section data has to be at the offset specified by the PointerToRawData field in the section header (see: https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#section-data).

This means that you can construct valid PE files that have data between section header and section data, loading these with the existing pattern would result in the added data being interpreted as section data.

WerWolv:
Thank you!

PR #303, opened 2024-09-26
patterns: Added pattern for Blender project files

Hikodroid:
Adds a pattern for Blender project files. The pattern gives a top-level overview of a blend file and is also able to visualize the embedded thumbnail, even if the file is somewhat incomplete. Note that the creation of the thumbnail section is imperfect due to this issue https://github.com/WerWolv/PatternLanguage/issues/130 and the section may be completely zeroed out if the file is truncated exactly after the end of the thumbnail block, but those issues do not seem to be caused by the pattern code.


paxcut:
Just a quick comment. Pattern language is able to decompress zstd compressed files  so maybe zstd compressed blender projects could be supported in the future.

WerWolv:
Thank you!

PR #304, opened 2024-09-26
patterns: Added pattern for LZNT1 compressed data

Hikodroid:
Adds a pattern for viewing and decompressing LZNT1 compressed data. LZNT1 is the default compression algorithm used by NTFS to optionally compress its data runs.


WerWolv:
Thank you!

PR #305, opened 2024-09-28
patterns/gb: Add 16K and 32K arrays for ROM Banks

bbbbbr:
Having arrays for dividing the GB ROM up into ROM Banks makes it much easier to jump to and inspect data at the start and end of banks.

- 16K Banks are the typical standard
- 32K banks aren't very common but are used in a couple unlicensed Games such as those using the Wisdom Tree MBC controller

imhex_gb-rombanks


WerWolv:
Thank you!

PR #307, opened 2024-10-02
Add Minecraft LCE ARC file format

DexrnZacAttack:
> [!NOTE]
> The example file is not originally from the game because I'm not sure if ripping directly from the game would cause issues.   

ARC is used as a container to store game and DLC assets, such as FUIs, minigame stuff, etc.   
I have tested a few different arc files and this pattern has worked with all of them.


DexrnZacAttack:
for some reason the unit test fails when I use add_virtual_file wtf
it works fine using master@d49d82e imhex

WerWolv:
Thank you!

PR #309, opened 2024-10-06
lcesave.hexpat: use Virtual Filesystem + ZLib support + auto endian detection

DexrnZacAttack:


patterns/lcesave.hexpat
```diff
@@ -1,17 +1,16 @@
...
-
-// TODO: re-add nbt pattern
-// TODO: clean up getFileType
+// NOTE: SPEGHETTI CODE!
 
 import std.string;
+import std.mem;
+import std.core;
+import hex.dec;
+#ifdef __IMHEX__
```
C3pa, line 11: Where does the `__IMHEX__` define come from?

paxcut: It is defined internally before pattern is evaluated. All patterns have it.

C3pa: DexrnZacAttack, why are you testing for this define?

paxcut: It isn't my place to reply since you asked the author directly, but note that the commit that adds the test for the define has "ifdefs to fix actions moment " as comment, so I suspect that it has to do with unit tests which wouldn't define the __IMHEX__ variable because tests don't use the runtime. Also it would make little sense to warn user during a test run which is supposed to run automatically.

C3pa: Thanks Paxcut.

DexrnZacAttack: > Thanks, Paxcut.
Just saw this thread, yes this is the case, as the actions stuff was failing without that check.

WerWolv:
Thank you!

PR #314, opened 2024-10-25
Add parsing of general purpose bit flag in ZIP pattern

targodan:
Add "GeneralPurposeBitFlags" bitfield definition to the ZIP pattern. The names are based on the official ZIP format specification, although I chose to name the fields closer to their purpose rather than their official name. This is intended to make reading the pattern and its output easier. Example: official name "Language encoding flag (EFS)" my name "filenameAndCommentAreUtf8" because this immediately explains what it does.

I chose not to implement the specifics of the "compressionOptions" as their meanings are specific to the compression method used, and it would make the pattern much more complicated for (in my opinion) little gain.

I also chose to unify the names of the general purpose bit filed in the LocalFileHeader and the CentralDirectoryFileHeader. Previously, they had different names, suggesting they might have different meaning. According to the official docs though, these fields have the exact same meaning.

Official docs: https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT

### DIFF
--- a/patterns/zip.hexpat
+++ b/patterns/zip.hexpat
@@ -157,17 +157,17 @@ enum CompressionMethod : u16 {
 };
 
 bitfield GeneralPurposeBitFlags {
-    bool encrypted : 1;
-    bool compressionOptions : 2;
-    bool crcAndSizesInCDAndDataDescriptor : 1;
-    bool enhancedDefalting : 1;
-    bool patchedData : 1;
-    bool strongEncryption : 1;
-    bool unused : 4;
-    bool filenameAndCommentAreUtf8 : 1;
-    bool reservedPKWARE_0 : 1;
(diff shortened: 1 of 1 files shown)


paxcut:
bit-fields that have size of one bit are implicitly boolean (either 1 or zero) so there is no need to explicitly declare them as bool. If, If for some reason, you insist on being redundant you still should remove the bool type from the entry which is described as compression options since it makes no sense for a boolean to have more than two states.

I believe there is a typo in the entry called `enhancedDefalting` which should probably be `enhancedDeflating`.

targodan:
Good point on the bool, I was unaware, that the type is optional. And good spot on the typo!

Should be fixed now.

WerWolv:
Thank you!

PR #315, opened 2024-11-08
Fix pe not working when ordinalTableRVA is zero

l12436:
Export table might not exist in pe exe. if  ordinalTableRVA is zero. pattern will failed to parse


WerWolv:
Thank you!

PR #323, opened 2024-11-24
Patterns/GBA Header

GekySan:
Hello, this PR adds:

- Game Boy Advance (GBA) hexpat.

No test files could be included due to copyright.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,8 @@ Everything will immediately show up in ImHex's Content Store and gets bundled wi
 | File System | | [`patterns/fs.hexpat`](patterns/fs.hexpat) | Drive File System |
 | FLAC | `audio/flac` | [`patterns/flac.hexpat`](patterns/flac.hexpat) | Free Lossless Audio Codec, FLAC Audio Format |
 | Flipper Zero Settings | | [`patterns/flipper_settings.hexpat`](patterns/flipper_settings.hexpat) | Flipper Zero Settings Files |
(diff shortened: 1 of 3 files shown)


Lemon4ksan:
There's a typo in line 139 'return _Unknown Licensee";_

GekySan:
> There's a typo in line 139 'return _Unknown Licensee";_

Hello, I don’t see any typo errors. Could you clarify where it is, please ?

paxcut:
I am pretty sure Lemon4san means `Licensee` which in this case is not a typo, but what else could it  be, `return`? `unknown`?  I don't think so. 

Among other meaning, a licensee is a person or group who is in possession of a license.

GekySan:
> I am pretty sure Lemon4san means `Licensee` which in this case is not a typo, but what else could it be, `return`? `unknown`? I don't think so.
> 
> Among other meaning, a licensee is a person or group who is in possession of a license.

Ah, but I think `Licensee` is correct, and anyway, as mentioned in the comment in the code, I copied what [WerWolv did in Game Boy](https://github.com/WerWolv/ImHex-Patterns/blob/b9f5f1668bff1651189c5cefd9cf1bbc355a50a5/patterns/gb.hexpat#L230).

paxcut:
I wasn't clear enough and for that I am sorry. My comment was meant to indicate that the above mentioned post incorrectly classified `Licensee` as a typo. That's an understandable mistake if English is not your mother tongue.

I added  the definition for the benefit of those who doubt it is the correct spelling.

Lemon4ksan:
Sorry, my bad

WerWolv:
Thank you!

Issue #258, opened 2024-06-06
Request: .rar file pattern


Devnol:
I was trying to look into a rar file and noticed that no pattern existed for it, whereas zip and 7z both have pattern files. I could try to do this myself someday, I'm just putting it out here for now if someone else would like to work on it in the meantime.

This is the spec for the RAR 5 format: https://www.rarlab.com/technote.htm
And this seems to be an archive of the entire set of RAR specs? https://www.loc.gov/preservation/digital/formats/fdd/fdd000450.shtml

targodan:
I've been looking for an excuse to dive deeper into the imhex pattern language for a while now. I think I'll take a poke at this. :)

PR #324, opened 2024-11-27
Add rar file pattern as requested in #258

targodan:
# Pattern

The RAR file format as defined in the official documentation (https://www.rarlab.com/technote.htm). This pattern was requested in #258 .

There are still some issues I'd like to resolve, but I'm not versed enough in the pattern language to do so. I need some help here.

Issue 1: I have multiple cases where they have used a variable-length integer with certain bits indicating certain things. In some cases I got away with just using a fixed size bitfield, because the fields effectively were fixed width (at least with the linux rar implementation). In case of line [295](https://github.com/WerWolv/ImHex-Patterns/compare/master...targodan:ImHex-Patterns:rar?expand=1#diff-f9b8c68e79a8e9daee86a316064b8d0580072b3770ea88384859a9baa5481e7cR295) I did observe the field having variable width. I know of no way to cast the computed value of a `vint` into a bitfield. Maybe I'm missing a feature of the pattern language or it simply isn't possible.

Issue 2: As stated in line [360](https://github.com/WerWolv/ImHex-Patterns/compare/master...targodan:ImHex-Patterns:rar?expand=1#diff-f9b8c68e79a8e9daee86a316064b8d0580072b3770ea88384859a9baa5481e7cR360) I'd prefer to read headers until I reach a specific header, rather than end of file. This would allow for successful extraction of embedded rar files. Again, I don't know how to do this. Sadly, there is no absolute archive size anywhere that I could use instead.

## Checklist
- [x] A test file for this pattern has been added to [/tests/patterns/test_data](/tests/patterns/test_data)

## PS

FYI: I'm not sure why, but Github didn't load the PR template, I copy-pasted it manually.


targodan:
Not sure why unit tests of other patterns suddenly fail, my `rar.hexpat` passed the unit tests. I did not modify any existing files except the README.md

WerWolv:
Thanks a lot! The CI errors were my fault, sorry. I restarted the CI again now and it should hopefully pass

WerWolv:
About your questions,

Issue 1: 
It's possible to write bitfield fields whose length depend on other fields or other values in the pattern. For example:
```cpp
bitfield Test {
    field1 : 3;
    field2 : field1;
};
``

Issue 2:
Using `while(std::mem::eof())` is fine in my opinion. If you want to stop the evaluation of the array sooner though, you can use a `break` statement inside one of the structs that are placed by that array and it will terminate the array immediately. That way you should be able to stop once you hit that header

targodan:
Thanks for the suggestions :)

**Issue 1:**

I don't think this approach will work. The vint implementation is a little more complex. The MSBit is basically a continue indicator. As long as that bit is set, we need to read another byte. Then for the actual integer value we have to mask out the continuation bits and do some shifting. This computed value now acts as a bitfield.

You can see the decoding implementation in [line 33](https://github.com/targodan/ImHex-Patterns/blob/f003b9638465fdd1145ed322cfca35c59fea4e0d/patterns/rar.hexpat#L33).

So the problem is that I have a computed value (currently managed via both a `transform` and a `format` attribute) and the resulting value needs to be interpreted as a bit field. Is this possible?

**Issue 2:**

That's useful, I think. :) Implemented it.

paxcut:
That document seems to be for versions 4 and app. The links to earlier versions are dead but with a bit of searching I found the version [2](https://github.com/ashang/libxad/blob/master/support/formats/RAR202.txt) and one of the version [3](https://github.com/php/pecl-file_formats-rar/blob/master/technote.txt) specs. there are no records for version 1.

WerWolv:
I don't know the spec super well but if it's too much work to implement full support right now, feel free to just add an `std::assert` to abort evaluation with a nice error message if the version isn't supported. There can always be improved upon that then :)

PR #326, opened 2024-12-01
patterns: Add STDFv4 pattern

andreasWallner:
# Pattern

The [Standard test data format](https://en.wikipedia.org/wiki/Standard_Test_Data_Format) is defined to keep track of testing data during chip manufacturing and used by almost all tester manufacturers. The official documentation can be found at http://www.kanwoda.com/wp-content/uploads/2015/05/std-spec.pdf.

I tested it with an example published by Roos (https://roos.com/docs/RBEH-AH4R8T?Open) but did not include that in the PR as they list it as "confidential and proprietary" (though I find "confidential" an interesting statement for smthg hosted publicly...).
I also tested it multiple internal files from other vendors, which I can't share. I'm checking with a colleague whether they can provide some small file that doesn't include any proprietary information but my hopes are not too high - so I opened the PR anyways. I hope this is still OK, would raise another PR if/when I get a file.

The one that could be improved here is the displaying of nibble fields, but I didn't figure out a nice way to do it, other then creating a section - which kind of explodes with the usual file sizes.
Also editing could be improved...

# Checklist
[x] A pattern for this format doesn't exist yet (or this PR improves the existing one)
[x] The new pattern has been added to the relevant table in the Readme
[x] The new pattern has a description pragma (#pragma description My pattern Description here)
[x] The pattern was associated with all relevant MIME types (using #pragma MIME mime-type in the source code)
[ ] A test file for this pattern has been added to [/tests/patterns/test_data](https://github.com/tests/patterns/test_data)

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ Everything will immediately show up in ImHex's Content Store and gets bundled wi
 | shp | | [`patterns/shp.hexpat`](patterns/shp.hexpat) | ESRI shape file |
 | shx | | [`patterns/shx.hexpat`](patterns/shx.hexpat) | ESRI index file |
 | SPIRV | | [`patterns/spirv.hexpat`](patterns/spirv.hexpat) | SPIR-V header and instructions |
+| STDF | | [`patterns/stdfv4.hexpat`](patterns/stdfv4.hexpat) | Standard test data format for IC testers |
 | STL | `model/stl` | [`patterns/stl.hexpat`](patterns/stl.hexpat) | STL 3D Model format |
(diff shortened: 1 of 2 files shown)


WerWolv:
Great work, thanks a lot! Having a test file would be great so we can ensure the pattern doesn't break in the future, but I understand that these files are often confidential. I'm happy to merge a future PR if you manage to get a file.

> The one that could be improved here is the displaying of nibble fields, but I didn't figure out a nice way to do it, other then creating a section - which kind of explodes with the usual file sizes.

For sub-byte fields you can easily use bitfields to display those values, without having to resort to sections: https://docs.werwolv.net/pattern-language/core-language/data-types#bitfields

WerWolv:
Hey, were you able to get a test file?

PR #327, opened 2024-12-01
Build system tweaks

BobSmun:
* Add top level CMakeLists.txt, to make it easier to use both independently and from within other projects

* Add a unit_test target, to attach all unit tests to


WerWolv:
Looks good, thank you!

PR #328, opened 2024-12-04
Added DPAPI Blob , DPAPI MasterKey  , CREDHIST pattern file & Updated README.md

5h4rrK:
### Merge Request Description

**DPAPI MasterKey, Blob and CREDHIST Pattern File**

**Key Features:**
1. **DPAPI MasterKey:**
   - Binary structure with service header and four slots.
   - Includes user's Master Key, local encryption key, local backup key (or CREDHIST GUID), and domain backup key.
   - FilePath: `C:\Users\<USER>\AppData\Roaming\Microsoft\Protect\<SID>`
   - Files are hidden; use `attrib -h -s` in Command Prompt to reveal.

2. **DPAPI Blob:**
   - Binary structure containing application's private data encrypted using DPAPI.

3. **CREDHIST:**
   - Password history file with a chain of older password hashes.
   - Each password change appends the old hash, encrypted with the new password.
   - FilePath: `C:\Users\<USER>\AppData\Roaming\Microsoft\Protect\`


WerWolv:
Thank you very much! Would it be possible to upload some test files for these?

5h4rrK:
Sure, I’ll add the test files for these.

PR #329, opened 2024-12-12
Added: BZip3 File Header Pattern

Sewer56:
This adds parsing of BZip3 as a pattern for ImHex [documented here](https://github.com/kspalaiologos/bzip3/tree/master/doc)

image

I started using ImHex... 2-3 hours ago; so please let me know if I'm missing anything. 🤞


Sewer56:
Note per PR:

- https://github.com/kspalaiologos/bzip3/pull/141#issuecomment-2537816794

This is the file format, not the frame format. Both formats have the same magic header; which may cause someone to confuse one for the other. (I certainly did)

WerWolv:
Thanks a lot! Pattern looks good to me, however you should use `import std.mem;` nowadays instead of ´#include`

Sewer56:
a-ok, done :clap:

WerWolv:
Thank you!

PR #330, opened 2024-12-13
Fixed: Incorrect field on branch in bzip3

Sewer56:
Hi there.
Noticed a small thing, since I started fuzz testing bzip3 this morning.
Turns out I was branching on the wrong field; made an accidental error, because testing tiny files (< 64 bit) is rare.
Oopsie. In any case; this trivial fix should do it.


WerWolv:
Thank you!

PR #333, opened 2024-12-18
Fix parsing of signed leb128

hexagonrecursion:
1. 0x80 0x40 is -8192 (-0x2000) in signed leb128, but the pattern language thinks it is 8192
2. 0xFF 0x3F is 8191 (0x1fff) in signed leb128, but the pattern language thinks it is -8193

The detection of the sign bit was wrong.


WerWolv:
Thank you!

PR #337, opened 2025-01-09
Add sup pattern file

miasinterestinglife:
This is a Pattern to mostly parse PGS Subtitles (in .sup format). I don't know what to do about the MIME Type, as PGS subtitles don't have one as far as i know.


WerWolv:
Hey, Thanks a lot!

Don't worry about the MIME type. If there isn't one, just leave it as it is. The `magic`pragma will take care of the detection.
If possible, could you upload a test file to `/tests/patterns/test_data` so we can run the unit tests on your pattern in the future? Also please add the pattern to the table in the readme and if you like add a `#pragma author <Your Name>`to the pattern so we can give credit to you in the UI :)

miasinterestinglife:
That shoud be it, right?

miasinterestinglife:
It would be possible to add decoding of the actual images at a later point, but it would be a huge undertaking

WerWolv:
Thanks! Looks good yeah :)

Having it decode the images would be really cool but by no means required in the first version. I'm happy to merge any future PRs improving upon it though!

PR #339, opened 2025-01-13
Add veado pattern

bake:
Hi, I added a minimal working pattern file and example for [veadotube](https://veado.tube) avatar files :)


WerWolv:
Thanks a lot! :)

PR #343, opened 2025-01-19
Add some improvements to the ISO pattern matching

xZise:
Primarily it parses all volume descriptors and not just the first one. To do that it always allocates enough padding to reach a length of 2 KiB.

I also tried adding support for the supplementary volume descriptors assuming that it is using Joliet. I'm not sure if that second part fits the code quality of this repository and I can separate it into another pull request if requested: It might contain other encodings depending in the contents of the escape sequences.

### DIFF
--- a/patterns/iso.hexpat
+++ b/patterns/iso.hexpat
@@ -97,6 +97,10 @@ struct DirectoryRecord {
   u8 fileNameLen;
   char fileName[fileNameLen];
   padding[$ % 2];
+  u8 remainingSize = $ - addressof(this);
+  if (recordSize > remainingSize) {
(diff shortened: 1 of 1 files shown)


WerWolv:
Hey, thanks a lot! Would it be possible to test somehow if the data is Joliet or not? I'm completely okay with having it in there but it shouldn't break the pattern for files that don't use it.

xZise:
I've added an additional check, although I'm not sure whether this is actually necessary. The only thing that could break would be the format, when some supplementary descriptor isn't using UCS-2/UTF-16.

WerWolv:
Okay, that's fine for me then. Thanks a lot!

PR #344, opened 2025-01-23
patterns: Fix incorrect condition in iso pattern

xZise:
Somehow the condition got copied to the wrong line in #343 and nucked another field.


WerWolv:
Thanks!

PR #350, opened 2025-02-09
patterns: Add RPM pattern

k0tran:
First attempt at implementing RPM package file format in pattern language. Would like your review and critique.

File spec: http://ftp.rpm.org/max-rpm/s1-rpm-file-format-rpm-file-format.html

Would also note that I decided not to patternize LZMA archive which is in this file too. Firstly it is an RPM and it already has a lot of patterns. Secondly LZMA should be separate pattern


k0tran:
Actually it not always LZMA + CPIO
There is tag payloadformat which states for example CPIO
There is tag payloadcompressor which states for example lzma or xz
There is also payloadflags which specify compress level
[RPM Source](https://github.com/rpm-software-management/rpm/blob/e426de95e665f8deedb8bf25958823326ac78e0d/docs/manual/tags.md?plain=1#L95)

But it is package contents anyway so I would propose just labeling it as that

WerWolv:
Thanks a lot! Looks good to me.
Could you also upload a test file so the CI can run tests on your pattern?

PR #352, opened 2025-02-25
patterns: Add Unreal Engine 3 UPK pattern

Vemmi:
Unfortunately, no test data added.
The UPKs I have on hand are from game files, and their filesizes are pretty large anyway.

Should otherwise be fine though.


WerWolv:
Thank you!

PR #354, opened 2025-03-02
patterns/ani: Improved compatibility with error size defined in ani files

xhsgg12302:
In the process of learning `pattern-language`, I selected [ani.hexpat](https://github.com/WerWolv/ImHex-Patterns/blob/master/patterns/ani.hexpat) file with relatively little content for testing and learning. At that time, I did not use the test data **ani.hexpat.ani** in this repo, but found a [1718926927_SizeAll.ani](https://sample-files-online.com/samples/ani) on the Internet, that is, renamed **ani.compatibility.hexpat.ani** in PR.

Use *ani.hexpat* to evaluate *ani.compatibility.hexpat.ani* and console output error, then check and find that the size defined by `u32 @0x04` in file -ani.compatibility.hexpat.ani- is incorrect. However, other software such as [open-ani-online-](https://codesandbox.io/p/sandbox/open-ani-online-v6g3j) can parse this ani file online, so the pattern file is modified to be compatible with the incorrectly defined size of the ani file format.


WerWolv:
Thank you!

PR #355, opened 2025-03-04
Add a TI83/84 pattern

libewa:
The pattern applies to various files created by TI Connect (CE).
I have only uploaded a test program file, but could upload examples of most data types.
The pattern only covers the file and content header, not the binary content (although this could be done).


libewa:
ah yes, can someone help me fix the checksum function?

PR #357, opened 2025-03-06
use e_shstrndx to get section names instead of guessing

### LINKED ISSUE #356: [BUG] Elf pattern doesn't properly handle section name in some cases
Image

[yt-remote.fap.zip](https://github.com/user-attachments/files/19105348/yt-remote.fap.zip)

Checking rust-produced elf file (the extension .fap is used by flipper zero) gives nonsense section names. Seems like it's choosing wrong segment to use as section name string table.

Shadlock0133:
fixes #356


playday3008 (approved):
Had bug where it will use `STRTAB` with debug symbols instead of sections names, those changes fixed it

WerWolv:
Thank you!

PR #358, opened 2025-03-11
patterns/partition_table: add basic MBR and GPT patterns

RadxaYuntian:
Both are conbined into a single pattern due to how GPT always has a pseudo MBR to maintain backward compatibility.

---

Created because I'm checking some strange vendor U-Boot error, where it reports CRC errors on GPT. I wonder if there is an easy way to call scripts to do complex processing, so I don't have to implement the crypto library in our DSL.

Though this might also result in our DSL to become an alternative implementation of `pip/cargo/npm` to manage the script dependencies :/ I guess I'll just use ImHex to copy the binary out for now.


WerWolv:
Thanks a lot! Please also add the pattern to the table in the Readme and (if possible) upload a small test file

WerWolv:
Thank you!

PR #362, opened 2025-03-13
fix: Set elf endianness globally

fneddy:
for some reason the `set_endian` in the elf struct does not take effect. big endian elf files are not parsed correctly. setting the endianness globally fixes the issue.


WerWolv:
Thank you!

PR #363, opened 2025-03-13
patterns/zip: Fix 0x5455 extended timestamp fields crash, improve date/time presentation

ctrlaltf2:
For the 0x5455 extra fields extension, the current pattern doesn't handle the central directory / local file difference. In the central directory, only the ModTime field is supported. I came across a dataset (https://data.fcc.gov/download/pub/uls/complete/vs_market.zip) that sets more than just the ModTime flag in the central directory header, which with the current parser, causes a crash. This PR fixes that and also formats the DOS times present in the local file and central directory headers by default.


WerWolv:
Thank you!

PR #367, opened 2025-03-18
Add ZIM pattern

cjdenio:
Adds a pattern for [ZIM](https://wiki.openzim.org/wiki/OpenZIM) files! Works on small and large files alike (with the caveat of only decoding the first 1,500 entries to avoid crashing on REALLY big files).


patterns/zim.hexpat
```diff
@@ -0,0 +1,79 @@
...
+} [[inline]];
+
+struct ClusterPointer {
+    Cluster *cluster: u64;
+} [[inline]];
+
+struct Entry {
+    u16 mimetype [[format("mimename")]];
+    u8 parameterlen [[hidden]];
+    char namespacee;
```
C3pa, line 58: namespace

cjdenio: `namespace` was a reserved keyword and I didn't have an alternative. suggestions welcome!

C3pa: Oh right 😄

Naheulf: If it's only to display in ImHex you can use
```cpp
char _namespace [[name("namespace")]];
```

WerWolv:
Thank you!

PR #368, opened 2025-03-19
patterns: Add ADTFDAT pattern

mheimlich:
Adds a pattern for [ADTFDAT](https://digitalwerk.gitlab.io/solutions/adtf_content/adtf_base/adtf_file_library) which is the default file format used by [ADTF](https://digitalwerk.gitlab.io/solutions/adtf_content/adtf_base/adtf_core/index.html). These usually contain a large number of chunks, so we cannot display all of them. When the array/pattern limit is hit, we switch to validating the chunk headers only and in case of an invalid chunk header yield only this header with some diagnostics.


WerWolv:
Thank you!

PR #369, opened 2025-03-22
patterns/zip: Handling padded extra fields

5ec1cff:
Some zip files have their extra fields padded at the end  (such as apks processed by [zipalign](https://cs.android.com/android/platform/superproject/main/+/main:build/make/tools/zipalign/ZipEntry.cpp;l=210;drc=9c843a66d11d85e1f69e944f1b37314d3e47aab1))


WerWolv:
Thank you!

PR #374, opened 2025-03-24
readme: fix table formatting

RadxaYuntian:
Somehow missed that there is another column when adding the new line.


WerWolv:
Thanks ^^

Issue #346, opened 2025-01-30
Protocol Buffers always flags the same field number


MaxwellBo:
Let's load a protoscope test file: https://github.com/protocolbuffers/protoscope/blob/main/testdata/message.pb

We detect field `114` here:

<img width="942" alt="Image" src="https://github.com/user-attachments/assets/e0118850-02ea-4800-acf8-b43bbacf3fe5" />

and here

<img width="942" alt="Image" src="https://github.com/user-attachments/assets/2429d619-deef-46fe-a09f-99a021e592da" />

and here

<img width="942" alt="Image" src="https://github.com/user-attachments/assets/e7cdbf69-7aaf-42a0-a1b2-2616c133aefb" />

Let's try a file of my own - this time it thinks every field is `8`.

<img width="956" alt="Image" src="https://github.com/user-attachments/assets/ee465c1f-0c49-4f44-bb89-24a43878c627" />

<img width="956" alt="Image" src="https://github.com/user-attachments/assets/17d9e6e7-e506-48a1-840e-902a8d2b84e6" />

<img width="956" alt="Image" src="https://github.com/user-attachments/assets/0e6bfb85-805d-477e-9f1a-9c303d3a8606" />

paxcut:
I can't imagine this pattern ever worked correctly. It uses global variables to display values that are calculated as the fie is read, but the values that are printed in the format functions will always be the last value the global takes. In the first example that value is 114 in your own example I don't really know by I venture that the last value that the global variable `field_value` takes is 8. An easy way to check this is to insert a print statement in the place where the global variable is being updated:
```cpp
struct Key {
    std::print("{:#x}",$);
    type::uLEB128 keyDec;
    field_number = u32(keyDec) >> 3;
    std::print("{} {}",field_number, u32(keyDec) >> 3);
    wire_type = u32(keyDec) & 7;
}[[format("format_key")]];
```
where i also print the address of `KeyDec` and removed the `[[sealed]]` attribute. That prints the correct values one expects to see and  if you scroll all the way to the bottom you will find the value it 'thinks' all the other ones are. 
It should be easy to fix this but I know nothing of protobuffers so I don't know what else could be wrong.

paxcut:
I think this is what the pattern should be doing. Can you test it to make sure the results are what you expect? If it does feel free to open a PR or I can do it if you prefer not to. Let me know one way or the other.
```cpp
#pragma author WerWolv
[...]

PR #378, opened 2025-03-29
Improve protobuf parsing.

glennhartmann:
Mostly this enables attempted recursive parsing of submessages. Note that it is inherently impossible to determine the underlying data type for LengthDelimited for sure, so this is a best-effort attempt. The user can disable recursive submessage parsing via Settings.

Other minor changes:
* added #pragma MIME and #pragma endian directives
* enabled UTF-8 display for LengthDelimited types
* added signed LEB128 display for Varint types (although this doesn't seem to be working on my test case)
* swapped if/else-if structure for match
* fail upon receiving unknown or unsupported WireType

Tested against https://github.com/glennhartmann/ImHex-Patterns/tree/proto-test-case


glennhartmann:
Ah, looks like there's a related open issue that I didn't see until just now: https://github.com/WerWolv/ImHex-Patterns/issues/346

WerWolv:
Thanks a lot!

PR #379, opened 2025-03-31
Add magic number for UF2 pattern

cjdenio:
Adds a `#pragma magic` from the [spec](https://github.com/microsoft/uf2#file-format) so UF2 files can be auto-detected. 

Screenshot 2025-03-31 at 3 06 06 PM


WerWolv:
Thank you!

PR #381, opened 2025-04-04
Add type.mmreg, dls and wav hexpat

rsp4jack:
This PR adds type.mmreg module, DLS (DownLoadable Sounds) amd (M)XMF ((Mobile) eXtensible Music Format).

wav.hexpat is also modified to use `type.mmreg`.


WerWolv:
Thanks a lot! Looks great for the most part. 
There's one issue though with the placement of the files. The libraries in the `includes` folder are meant to be generic and to be included by many different patterns. So pattern specific data types don't really belong in there. Instead, since your patterns depend on multiple files, they should probably all go in a new sub-directory in the patterns folder so they can be downloaded and used all at once

rsp4jack:
@WerWolv the mmreg.pat is ported from mmreg.h in Windows SDK, standing for "Multimedia Registry". It contains the wFormat tags, WAVEFORMAT structs and something else.
Actually it is quite generic, but your suggestions make sense.

WerWolv:
Yeah, we don't have a proper way right now to have types and constants stored in some shared place. Maybe ultimately a new library would be good for that.
The types library, in any case, is meant for really standard types that a lot of different formats might use. Think types for parsing IP addresses, GUIDs, Fixed Point numbers and so on

PR #386, opened 2025-04-17
MachO Load Commands should include additional specializations (LoadDyLib)

DhashS:
Hi!

I've been banging my head looking for something like ImHex for something I'm working on. You've built something awesome, and I really like your pattern language.

I'd like to continue to flesh out the MachO (and maybe other) patterns, but I wanted to get this in so that I can check with maintainer that it's idiomatic / ask for advice

one thing that is strange is this line 
https://github.com/DhashS/ImHex-Patterns/blob/94da07d6955054bbd6f0798c7ef67a0972e83c7c/patterns/macho.hexpat#L398

since the null-terminated string is aligned to the nearest 8 bytes.

If this gets merged i'll continue ploughing through some of the symbol table


WerWolv:
Hey! Thanks a lot :)
If you just need to align the cursor to the next 8-Byte boundary, you can use something like `std::mem::AlignTo<8>;` instead of your manual calculation there. Alternatively, I'd use `padding[Your Calculation];` to just skip a few bytes instead of manually manipulating the cursor.

Besides that, `lc_str` feels a bit weird. Why is it a union? Just a regular struct would be more standard and does the same in this case

DhashS:
sweet. 

I found `padding` but felt weird that the index wasn’t a constant.

The single element struct is an error on my part, that union is a pointer on 32-bit systems

DhashS:
@WerWolv PTAL

DhashS:
@WerWolv I have more patches ready for MachO, but I don't want to put in the PR just yet in case something is stylistically wrong with what I'm doing. Will submit when this PR is merged.

PR #387, opened 2025-04-17
patterns: Add definition for the OpenRGB profile format

DEREFERENC3D:
Adds pattern definition for the binary OpenRGB profile format (`.orp` / `.ors`). Supports up to version 5 - latest as of the current OpenRGB `master` branch.


WerWolv:
Thank you!

PR #390, opened 2025-04-27
patterns: Add support for MagicaVoxel .vox files

JuanDiegoMontoya:
Adds support for MagicaVoxel scene description files (.vox).

References:

- https://github.com/ephtracy/voxel-model/blob/master/MagicaVoxel-file-format-vox.txt
- https://github.com/ephtracy/voxel-model/blob/master/MagicaVoxel-file-format-vox-extension.txt

Test files can be found [here](https://github.com/ephtracy/voxel-model) and in the `vox` subfolder of the MagicaVoxel installation directory.


paxcut:
Thanks for this. Just have a couple of questions/comments.

I see you are defining structs for RGBA8 and String. Any reason why not using the ones in the libraries? 

If you look in the libimhex repo code in file include/hex/helpers/opengl.hpp you'll find function getRotationMatrix that may be what you have in the TODO and has extensive documentation.

Any chance of providing a small ( kb range  at most) file that can be used for unit tests to make sure that changes don't break the pattern in the future?

JuanDiegoMontoya:
Hey @paxcut.

> I see you are defining structs for RGBA8 and String. Any reason why not using the ones in the libraries?

The reason I defined my own RGBA8 and String structs was to ensure they had the same layouts as the ones that appear in .vox files. That reasoning is dubious for RGBA8, so I tried replacing it with the one in `type`, but couldn't figure out how to make its color appear in the Pattern Data window. They're all blue, which, after scanning the Discord, seems like a problem you've encountered as well.

```cc
using RGBA8 = type::RGBA8 [[color(std::format("{:02X}{:02X}{:02X}", r, g, b))]];
```

Since my RGBA8 struct is only a few more lines than this, I'll keep it for now.

> If you look in the libimhex repo code in file include/hex/helpers/opengl.hpp you'll find function getRotationMatrix that may be what you have in the TODO and has extensive documentation.

The reason the rotation function is not included is because I couldn't think of a good way to include it as a visualization for the Rotation type. I'll see if I can fix that- decoding isn't too difficult.

> Any chance of providing a small ( kb range at most) file that can be used for unit tests to make sure that changes don't break the pattern in the future?

Sure, I'll make a small model to test it out. I'm not very skilled with MagicaVoxel but I'll see what I can do 😄. FYI, the smallest that MagicaVoxel can export is in the tens of KBs due to the material info it emits. Hopefully that isn't an issue.

JuanDiegoMontoya:
I added a test file to the commit. It should be pretty much ready to go.

The only thing I didn't do was a fancy formatter with the RotationToMat3 function because it was turning out to be too complicated for what it would add (rotations rarely show up in vox files).

paxcut:
The rgba color in the pattern editor was fixed already so there is no need to create an identical version of the type. There are ways to define types using the existing  library types so that you can add your own formatting functions to what is essentially the library type. The string struct could be using that since it is only different in the format function.

 I don't mean to make a big deal out of this and code duplication is always present but i see no reason not to mention it when i see it either.

I know voxel models are fundamentally different from vertex based models. Are there methods to obtain meshes from them? maybe it is possible to use the 3d visualizer to display some approximation to the model. This is just me thinking out loud, so disregard if it is nonsense. 

I only mentioned the rotation matrix because i saw your todo comment and not as something that needs to be worked on. I m sorry if it seemed that way.

JuanDiegoMontoya:
Yeah, code duplication isn't ideal. If it's a blocker then I'll go back and remove those types.

MagicaVoxel can export polygon meshes from these models in a few common formats, and I bet some other programs can too. Otherwise, you'd have to write code to extract a mesh from them (they could be treated as simple cubes or a smooth isosurface for an algorithm like marching cubes or surface nets).

The rotation matrix thing is alright, someone else could use the code I wrote for learning or for enhancing the pattern. I can hardly stand to see a lone TODO comment anyway.

PR #391, opened 2025-04-29
added NTFS pattern

HrantTadevosyan:
- [x] Tested on 4k cluster size, NTFS volumes
- [x] Tested on 8k cluster size, NTFS volumes

## For Reviewers

- This will parse the `VBR` and known first `MFT` records
- Additional sample pointers are parsed for `$AttrDef` and `I30 (Root Directory)`
- Majority of the attributes will be parsed automatically
- I used `ntfs-3g` driver as a reference point, since to me it looked like the most sophisticated place

what is missing:

- [ ] Parsing all attribute types and their corresponding bodies
- [ ] Iterating over BTrees
- [ ] Sanity checks... a lot of them (magic number checks, parent vs child(ren) size checks, preliminary type checks, etc).
- [ ] ...

and probably a whole  lot

## References

- https://github.com/libyal/libfsntfs
- https://github.com/tuxera/ntfs-3g


paxcut:
is it possible to run the pattern on a sample input file? If so, do you have one that is small that can be added to the nit tests?
The pattern itself looks good but I see lots of definitions for le and be variables when it looks like it is mostly the default le that is used. Why not simply use the builtin types.? Although generally small, the typedef types do add a little overhead over using the builtin types directly. In some cases, the overhead can become quite large as was seen in one pattern that was using gbs of ram instead of mbs because of typedefs. Also you seem to be defining signed integers as unsigned. Is that intended ?

HrantTadevosyan:
These typedefs are there to match `ntfs-3g` definitions. I wanted to keep it as close to original as possible. Unfortunately I ended up changing a lot of stuff....

- `types.h`: https://github.com/tuxera/ntfs-3g/blob/edge/include/ntfs-3g/types.h
- `layout.h`: https://github.com/tuxera/ntfs-3g/blob/edge/include/ntfs-3g/layout.h

About unit tests: My samples are usually `1G+` in size. Not sure how convenient they're for this repo. In future maybe I'll manually build a sample data, instead of relying on Windows [FORMAT](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/format) command.

paxcut:
In the definitions you posted they define the same type as both le and be but you are not doing that.They mention be architectures probably ARM is one of them but there doesn't seem to be any check for architecture so im a bit confused as to how that works.  I would try running the code without all the using statements. If it was me I would comment out the ones that are never used and the ones that are used try using defines instead and see if it affects time of pattern evaluation or memory usage specially if you are running on gbs worth of input files.

Yeah we ask for small test input files in the kb size range if possible. If not it is no big deal .

HrantTadevosyan:
They don't use `le u64` and `be u64` because C does not have such keywords... they use it as a 'decoration'. The values are correctly parsed by these macros:
- `endians.h`: https://github.com/tuxera/ntfs-3g/blob/edge/include/ntfs-3g/endians.h

Either way, I like the idea of using `#define`s. I'll update the PR, I think.

paxcut:
Yes that makes more sense now. Then I recommend not using `le` and `be` and just setting the endianess globally. `le` and `be` are only used when you need a few variables of the opposite endianess, not as a way to define variables of a particular endianess. For that you use a pragma to set it globally and the you use the built in types directly like u32 or s64. you can use define to give them the names you want so you can keep using the structs provided by the source code.

HrantTadevosyan:
Ok, I have push new changes. These should include:
- Improved index entries parsing
- Improved attributes parsing
- Added a new `2M` test data
    - The test data includes a test file, called `FINDME.txt`, which has the text `YOU FOUND ME`
- Removed `le64`, `be64`... stuff

paxcut:
yes, im afraid 2 mb test file is too big. A few kb is best.Keep in mind the large number of patterns in the repo so tests cant be too big. just remove it for now and you can always add one later if you  find something with suitable size. Other than that it looks good to merge.

HrantTadevosyan:
- removed the testdata
- made `$SECURITY_DESCRIPTOR` more generic

paxcut:
Thank you for that. I'll be merging as soon as the tests finish so if you have last minute changes nows the time.

PR #392, opened 2025-05-03
Added DJI Firmware Pattern

0x0000z3r0:
- [x] Tested on DJI Config Files
- [x] Tested on DJI Firmware Files

## For Reviewers

- Added DJI firmware structures. The definitions are not documented and were reverse engineered by the community (check `refs` in comments)
- Added a cropped sample file. It contains enough information for headers, but does not contain the entire payload (to make the test data small).


WerWolv:
Thank you!

PR #393, opened 2025-05-07
Fixes for bencode pattern

Niketin:
This pull request contains several fixes for the existing Bencode pattern, and also a fix for the std::ctype pattern.


WerWolv:
Thank you!

PR #397, opened 2025-05-16
pattern wintec_tes: slightly nicer formatting in a list of entries

ccoenen:
When viewing a wintec file, it typically consists of hundreds of entries line by line. I found out that any entry can be formatted, so the lines themselves now also get a readable string. This now looks like this:

image

(it used to display just `{ ... }` in the Value-column)


WerWolv:
Thanks a lot :)

PR #399, opened 2025-05-17
patterns: Add support for smk

xZise:
While working on another project I needed to decode smacker files and created this pattern.

Unfortunately I don't have any smacker file where I know that there is no copyright. Testing with a few files from https://samples.ffmpeg.org/game-formats/smacker/ works (except for `mech2/mintro.smk` which also won't work in VLC).


patterns/smk.hexpat
```diff
@@ -0,0 +1,216 @@
...
+
+struct AudioTrack<auto trackIndex> {
+    u32 length;
+    if (header.AudioRate[trackIndex].isCompressed) {
+        u32 decompressedSize;
+    }
+    u8 trackData[length - ($ - addressof(this))];
+};
+
+u32 frame_index = 0;
```
paxcut, line 180: using a global for frame_index is not necessary because you have std::core::array_index() for that exact use. Whenever you create an array, this function is available in any of the structs being used as elements in the array and it will return the index of the current element of the array being added. For example:
```cpp
// remove the global!
struct FramesData {
      u32 frame_index = std::core::array_index();
       u32 size = sizes[frame_index].size;
    if (frameTypes[frame_index].ContainsPaletteRecord) {
        PaletteChunk palette;
    }
    .
    .
    .
};
```
Then instead of checking the global variable to create the array of frames you can simply do:
```cpp
    FramesData frames[header.FrameCount];
```
Below I use that inside the container for the pattern

patterns/smk.hexpat
```diff
@@ -0,0 +1,216 @@
...
+    }
+    if (frameTypes[frame_index].ContainsAudioTrack6) {
+        AudioTrack<6> audioTrack6;
+    }
+    u8 video[size - ($ - addressof(this))];
+
+    frame_index = frame_index + 1;
+};
+
+Header header @ $;
```
paxcut, line 212: Wouldn't it be better if all these placements could be made in just one? You can create a container struct for the entire smk pattern. Something like
```cpp
struct SMK     {
    Header header;
    Size sizes[header.FrameCount];
    FrameType frameTypes[header.FrameCount] ;
    u8 trees[header.TreesSize];
    FramesData frames[header.FrameCount];
};

SMK smk@0x00 [[inline]];
```
using inline ensures that the resulting patterns look the same in the pattern data window but you'll be able to import smk pattern as a module with the name that you give it in the import * statement. Placing many patterns will make the module importer use variable names from the pattern instead of the custom ones.

paxcut (changes requested):
Hello, This looks very good. I am proposing a couple of changes to make it a little bit better by removing an unnecessary global and using a builtin function instead and by enabling importing this pattern as a module by putting all the placed patterns inside a single container and placing that container. This won't add an extra layer to the tree thanks to the [[inline]] attribute.

Let me know what you think.

paxcut (approved):
This looks very good. Going forward.

PR #400, opened 2025-05-17
added APFS pattern

HrantTadevosyan:
- [x] Tested on 1G+ samples

## For Reviewers

This should parse the container, volume superblocks. It should be able to identify the root, snapshot trees, the user can iterate over the B+Trees with helper functions. Individual blocks can be verified as well with these helper functions. Additionally, there are space manager, reaper and bitmap parsed structures. Lastly, it will try to find the root directory of the first volume in the container.

Unfortunately, I can not add <1M sample file for unit tests, since I don't know how'd I do it. The tools that I use allow only 1G+ disks to be created.

## References
- https://developer.apple.com/support/downloads/Apple-File-System-Reference.pdf
- https://github.com/sgan81/apfs-fuse
- https://github.com/libyal/libfsapfs


paxcut (changes requested):
I noticed a possible runtime error in the pattern. I posted details and a possible resolution.

HrantTadevosyan:
- I use VirtualBuddy, which does not allow me to create <1G disks. But... I see this issue in their repo:
https://github.com/netbox-community/netbox/issues/14692
Most likely I have to update my old VirtualBuddy.... in future. c;

- Oh that's a very good catch. I have never seen that error on my sample disks. But good point. I'll fix it once I come back to this. I'll tag you, once I push my changes (for you to double check).

WerWolv:
Hey, is this ready for merging or would you like me to wait?

PR #401, opened 2025-05-23
Add flv.hexpat

zn123:
https://veovera.org/docs/enhanced/enhanced-rtmp-v2#flv-file-format-overview

### DIFF
--- a/patterns/flv.hexpat
+++ b/patterns/flv.hexpat
@@ -1,0 +1,229 @@
+#pragma author ZN123 version 1.0
+#pragma description flv Archive
+#pragma MIME video/x-flv
+#pragma magic [46 4C 56] @ 0x00
+
+// zn123@sina.com
+// https://veovera.org/docs/enhanced/enhanced-rtmp-v2#flv-file-format-overview
+
+import std.io;
+import type.base;
+import std.mem;
+import std.math;
+import std.string;
+#pragma endian big
+#pragma pattern_limit 2000000
+
+enum TypeB:u24{
+    flvSignature = 0x564C46,  // Defining flv signature
+};
+#define FLV_HEADER_FLAG_HASVIDEO  1
+#define FLV_HEADER_FLAG_HASAUDIO  4
+struct FLVHeader {
+    char signature[3]  [[color("FF0000")] ];
+    u8 version [[color("00FF00")]];
+    u8 stream_info [[color("0000FF")]]; //si->missing_streams = flags & (FLV_HEADER_FLAG_HASVIDEO | FLV_HEADER_FLAG_HASAUDIO);
+	if (stream_info & FLV_HEADER_FLAG_HASVIDEO)
+		std::print("HAS VIDEO");
+	if (stream_info & FLV_HEADER_FLAG_HASAUDIO)
+		std::print("HAS AUDIO");
+    u32 head_size [[color("FFFF00")]];
+	u32 previous_tag_size [[comment("away 0")]];
+};
+enum FLV_TAG_TYPE : u8 {
+  FLV_TAG_TYPE_AUDIO = 8,
+  FLV_TAG_TYPE_VIDEO = 9,
+  FLV_TAG_TYPE_META = 18
+};
+
+#define FLV_AUDIO_CODECID_MASK 0xf0
+
+#define FLV_VIDEO_CODECID_MASK    0x0f
+#define FLV_VIDEO_FRAMETYPE_MASK  0x70
+#define FLV_VIDEO_FRAMETYPE_OFFSET   4
+
+
+
+
+//  type = (avio_r8(s->pb) & 0x1F);
+//  dts  = avio_rb24(s->pb);
+//  dts |= (unsigned)avio_r8(s->pb) << 24;
+struct BaseTag{
+	// if (std::mem::eof()) break;
+	u8	TagType;
+	u24 offset  [[comment("offset+11 = taglen")]];
+	u24 Timestamp;
+	u8  TimestampExtended;
+	u24 StreamID [[comment("away 0")]];
+	// std::print("TagType= {}",TagType & 0x1F );
+	// std::print("offset= {}",offset + 11);
+	// std::print("dts={},{}, {}",Timestamp,TimestampExtended,Timestamp + (TimestampExtended<<3));
+};
+enum AMFDataType:u8 {
+    AMF_DATA_TYPE_NUMBER      = 0x00,
+    AMF_DATA_TYPE_BOOL        = 0x01,
+    AMF_DATA_TYPE_STRING      = 0x02,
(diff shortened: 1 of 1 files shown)


zn123:
[test.zip](https://github.com/user-attachments/files/20433799/test.zip)

zn123:
> 你测试过自动检测模式文件的魔法吗？ 你可能需要设置一下，这样只需要在末尾放置一个变量，而不是两个。这样导入会更简单。

Consider the compatibility of different versions

zn123:
@reviewer ：Could you please help expedite the review process for this PR when you have a moment? Thank you!

WerWolv:
Please upload the test file to the folder mentioned in the PR template and add your pattern to the table in the readme. After that it should be good to merge 👍

zn123:
> 请将测试文件上传到 PR 模板中提到的文件夹，并将你的模式添加到 readme 文件中的表格中。之后就可以合并了👍

already submitted

patterns/flv.hexpat
```diff
@@ -0,0 +1,229 @@
...
+		u32 previous_tag_size;
+	} catch {
+        std::warning("Invalid sample payload at chunk");
+		break;
+	}
+} [[name(std::format("FLVTag({})", TagType))]];
+
+
+FLVHeader flvheader @ 00;
+FLVTag flvtag[while(!std::mem::eof())] @ $ ;
```
paxcut, line 229: This is what I meant by putting all variables in one placement. I don't understand your reply about considering different versions. What I suggested is that instead of placing `flvheader` and `flvtag` that maybe you want to create and FLV struct like this:
```cpp
struct FLV {
    FLVHeader flvheader;
   FLVTag flvtag[while(!std::mem::eof())];
};

FLV flv@0;
```
as you see only one variable is placed. if you want to use import * on the flv pattern then this way it is eadier to use than if you set two variables.

patterns/flv.hexpat
```diff
@@ -0,0 +1,229 @@
+#pragma author ZN123 version 1.0
+#pragma description flv Archive
+#pragma MIME video/x-flv
+#pragma magic [46 4C 56] @ 0x00
```
paxcut, line 4: This may not work. Last time I checked pragma magic required spaces around square brackets like in `[ 46 4C 56 ]` which is why I asked you if you have tested the auto loading in ImHex. Did you run the test I asked? Did the pattern auto load? If it works without spaces now I would like to know.

paxcut (commented):
this looks good. Only one possible problem and one suggestion that would make using this pattern easier.

patterns/flv.hexpat
```diff
@@ -0,0 +1,229 @@
...
+    u8 version [[color("00FF00")]];
+    u8 stream_info [[color("0000FF")]]; //si->missing_streams = flags & (FLV_HEADER_FLAG_HASVIDEO | FLV_HEADER_FLAG_HASAUDIO);
+	if (stream_info & FLV_HEADER_FLAG_HASVIDEO)
+		std::print("HAS VIDEO");
+	if (stream_info & FLV_HEADER_FLAG_HASAUDIO)
+		std::print("HAS AUDIO");
+    u32 head_size [[color("FFFF00")]];
+	u32 previous_tag_size [[comment("away 0")]];
+};
+enum FLV_TAG_TYPE : u8 {
```
paxcut, line 33: Your test data has only 3 types of tags audio video and meta but the variables are called `FLVTag(8)`, `FLVTag(9)` and `FLVTag(18)` that name means nothing unless you know what the numbers mean. Wouldn't it be better to call the variables `FLVTagAudio`, `FLVTagVideo` and `FLVTagMeta`?

patterns/flv.hexpat
```diff
@@ -0,0 +1,229 @@
...
+    } else if (std::string::contains(name, "onCaptionInfo")) {
+		std::print("onCaptionInfo");
+    } else if (std::string::contains(name, "onTextData")) {
+		std::print("onTextData");
+    } else {
+        std::error("Invalid DataEntryBox");
+    }
+	return flags;
+};
+/***
```
paxcut, line 114: A lot of code is commented out. If the code is not needed or only used in debugging keep it like that in your local version but the one posted to the pattern repo is the one others may be using and seeing all those comments doesn't help figure how the pattern works. Also you are using doc comments which are used to add documentation but you are not adding documentation . a doc comment starts with `/**`. a single line doc comment starts with `///`. I found an empty function that is never used. The pattern could use some cleaning up.

To help other read the pattern easier put empty lines between struct enums and bitfields to help indicate where they begin and end. These are only suggestions since they don't affect how the pattern runs but posting presentable code to a public repository is just a nice thing to do.

patterns/flv.hexpat
zn123: > 这就是我所说的将所有变量放在一个位置。我不明白你关于考虑不同版本的回复。我的建议是，与其放置`flvheader`和，`flvtag`不如创建一个像这样的 FLV 结构体：
> 
> ```c++
> struct FLV {
>     FLVHeader flvheader;
>    FLVTag flvtag[while(!std::mem::eof())];
> };
> 
> FLV flv@0;
> ```
> 
> 如您所见，仅放置了一个变量。如果您想在 flv 模式上使用 import *，那么这种方式比设置两个变量更容易使用。

Thank you for your suggestion, but this way adds another layer and doesn't bring any additional benefits. I hope the structure can be made as expansive as possible.

patterns/flv.hexpat
zn123: This is to inform users of the possible values of this position, for the convenience of learning; the annotations are close to the values, making it easy to view within one screen; I have considered your question and deliberately designed it this way.

patterns/flv.hexpat
paxcut: I did explain what the benefit is but I will explain again and with more detail. It is now possible to import patterns as modules with the import * syntax. That allows you to place the pattern at an arbitrary location within the importing pattern. If more than one variable is placed, then they will all be placed but not under the alias name given in the import command.

Attributes like [[flatten]] and [[inline]] can be used to eliminate any layer you want. Single placement is a common theme among all the patterns in the pattern language library and for good reasons. Easier maintenance, debugging, access to data, and the ability to use the entire pattern functionality with just one command. 

Understand that this is only a suggestion though and by no means you need to follow it or even explain why you don't want to follow it.

zn123: 接受建议

zn123: accept suggestions

patterns/flv.hexpat
paxcut: Ok, as long as you have considered them that's all i ask. thank you.

paxcut (approved):
Thanks for everything. I am approving this pattern and will merge when tests are passed.

PR #402, opened 2025-05-25
Quake 3 Engine demo pattern & test file

TomArrow:
Pretty basic format. The data in the messages themselves probably is too complex (not sure). It's huffman-data with a fixed dictionary, but it can contain individual bits too, so to my knowledge data must be read in the correct order based on complex logic that can vary between games. But at least with this individual messages and their lengths can be seen.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ Everything will immediately show up in ImHex's Content Store and gets bundled wi
 | PYC | `application/x-bytecode.python` | [`patterns/pyc.hexpat`](patterns/pyc.hexpat) | Python bytecode files |
 | QBCL | | [`patterns/qbcl.hexpat`](patterns/qbcl.hexpat) | Qubicle voxel scene project file |
 | QOI | `image/qoi` | [`patterns/qoi.hexpat`](patterns/qoi.hexpat) | QOI image files |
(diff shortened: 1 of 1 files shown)


TomArrow:
Figured out how to at least detect the type of the message. Might be possible to do a bit more for gamestates (read server hostname and such), but that would only be practical if I could pass the data array to some function to loop through strings.

TomArrow:
Okay added a bunch more features after I figured out how to use functions. However it's gotten quite slow compared to the original version, to the point where processing a 20 MB file takes a few minutes. Would it be prudent to create a lighter version as a variant? Or what's the usual approach with such things?

TomArrow:
Probably not gonna do anything more to this as it seems to work fine, unless someone finds some problems.

paxcut:
ImHex doesn't understand extensions and doesnt use them for anything. The column headers in the readme file are 
Name,  MIME,  Path  and  Description. dm_** is not a good name maybe "quake demo" pr something along those lines.The name is not really too important but the mime and the description are what defines what the format that the pattern exposes is.

paxcut:
Yes that looks good.thank you. I am ready to approve this then.

PR #406, opened 2025-06-05
Add PEF (Preffered Executable Format) support

domin568:
During implementing PEF support for "object" Rust crate I created pattern for PEF files. It is old executable format used by 68k/PowerPC Mac OS 7,8,9 but also supported up to OSX 10.4. There were some games that were compiled to PEF instead of Macho in late PowerPC Apple era. That's why I started getting into that format.

Pattern in PR shows sections, imports, exports, maps name offsets to proper strings. Relocation opcodes are also parsed. For samples I have it works just fine. If it is expected I can compile some more code and upload to this PR some test files.


paxcut:
Yes, if you could add a test file (small no mb sizes) that would be perfect but it isn't mandatory. Let me know.

paxcut:
Great, thanks for that. this is ready to go now.

paxcut (approved):
It all looks good.Thank you.

PR #407, opened 2025-06-11
Docs: describe how to create multidimensional `std::Array`

C3pa:
Fixes: https://github.com/WerWolv/ImHex/issues/1789


paxcut:
I think the reason the author asked for the example to be added was because the error message he got was confusing. instead of `No variable named 'dim1' found` a better error message would have been to state that `No variable named 'dim1' found in the scope where 'dim1' is used`. The first one makes you question how variables are being searched, but the second one explains that this is a scope issue which makes the resolution much easier.

But its too late for that. The example looks good but I am not sure what it accomplishes. It seems like another random example in a documentation that needs many more examples than the ones provided. 

Sorry for the rant. I'll approve and merge this now.

paxcut (approved):
Looks good and ready to go.

PR #409, opened 2025-06-17
Add pattern for Vector BLF Frame Logging Format

spnda:
Adds a basic pattern for the Binary Logging Format used by VECTOR Informatik GmbH for logging various protocols found in vehicles, such as CAN (FD), LIN, MOST, FlexRay, Ethernet, and some others. This pattern currently only supports standard CAN and CAN FD frames, and the object structs would need to be added for the rest.

Note that the format (or the official exporter) is slightly stupid and the pattern might not work for other files with other data, since for example for `app_text` object the object size falsely does not include the null terminator of the string, making it impossible to "just skip" the object in the object array.

I've opened this as a draft for now, since I'd be interested in getting some general feedback of how this pattern is written since this is my first ever, and I might look into adding more object formats in the future. There's also plenty of things I am not sure about how they work, and there's a lot of "padding" here that might actually contain data that I just don't know about. However, I would like for this format to be in this repository so that other people who might need to work with this format have a good reference as to how it is laid out.


patterns/blf.hexpat
```diff
@@ -0,0 +1,535 @@
...
+    u64 timestamp;
+};
+struct obj_v2_header {
+    object_flags flags;
+    timestamp_status timestamp_status;
+    padding[1];
+    u16 object_version;
+    u64 object_timestamp;
+    u64 original_timestamp;
+};
```
paxcut, line 395: You could combine both header versions into one
```cpp
struct obj_header {
     object_flags flags;
     if (parent.header.header_version == 1) 
         u16 client_index;
    else if (parent.header.header_version == 2) {
         timestamp_status timestamp_status;
        padding[1];
    }
    u16 object_version;
    u64 timestamp;
    if (parent.header.header_version == 2)
        u64 original_timestamp;
};
```
then in obj_struct:
```cpp
    obj_header header [[inline]];
```

PR #411, opened 2025-06-19
feat(pcapng): add  support for Decryption Secrets Block

ruokeqx:


patterns/pcapng.hexpat
```diff
@@ -410,12 +414,16 @@ struct CompressionBlock{
...
 
-// Experimental
 struct EncryptionBlock{
     BlockType block_type[[name("BlockType")]]; 
     u32 block_len1[[name("BlockLen1")]];  
     EncryptionType comp_type [[name("Encryption")]];
-    char data[block_len1];
+    u32 secrets_len[[name("SecretsLength")]];  
+    char data[secrets_len];
+    $ = $ + (4 - ($ % 4 ) ); // Padding
```
paxcut, line 423: if the purpose of this padding is to ensure that data starts at a 4 byte boundary then this will not work. If `$` is already divisible by 4 then it shouldn't change at all, but this will make it advance by 4. 

What you need is an expression that for `$` in `[0..N]` it returns the sequence `[0,3,2,1,0,3,2,1,0,3,2,1,....]`. Instead of manipulating `$` directly it is recommended to use the `padding` built-in data type. One way to obtain the desired sequence would be:
```cpp
padding[ ((4 - ($ % 4)) % 4)];
```
the last modulo makes sure that $ doesn't change if `$` is already at a 4-byte boundary. 

The expression above can be further simplified when N is a power of 2 using `x % N=x & (N-1)`. For N=4 you can obtain the same effect using:
```cpp
padding[-$ & 3];
```
which is much simpler and faster to calculate. The - sign ensures that instead of getting the sequence `[0,1,2,3,0,1,2,3,...]` you get the correct one from above.

paxcut (changes requested):
I think there may be a problem with the padding to get aligned data. Pattern language provides an aligned type that may do what you need, but I also wrote down a way to calculate it manually if you prefer to do that. The important thing is that no padding should occur if the offset is a multiple of 4 already.

patterns/pcapng.hexpat
ruokeqx: thanks for the tip.

ruokeqx:
> I think there may be a problem with the padding to get aligned data. Pattern language provides an aligned type that may do what you need, but I also wrote down a way to calculate it manually if you prefer to do that. The important thing is that no padding should occur if the offset is a multiple of 4 already.

fixed. thanks

paxcut (approved):
thanks for the fix. Everything else looks good.

PR #412, opened 2025-06-22
patterns: Add json support for glb files

xZise:
This makes it possible to separate display the different buffer views, accessors and images (even visualizing them).

Unfortunately the data within the JSON gets sometimes corrupted and this is the reason, why it parses the JSON multiple times at some places.


xZise:
I've uploaded this as a draft, because it parses the JSON multiple times and preferably this shouldn't be necessary. I'm also not sure with what formatting it should work. Newer files seem to use UpperCamelCase for types and snake_case for fields. Currently the original part uses the C++ style with the `_t` suffix.

It also fails the unit test as the `Json` type is not available without ImHex.

patterns/gltf.hexpat
```diff
@@ -53,7 +56,10 @@ enum gltf_chunk_type_t : u32 {
 struct gltf_chunk_t {
     u32 length;             /**< Length of this chunk. */
     gltf_chunk_type_t type [[format("gltf_format")]]; /**< Type of the chunk. JSON or BIN expected. */
-    u8 string[length];      /**< The chunk data. */
```
paxcut, line 56: This is one way you could pass unit tests.
```cpp
#ifndef __IMHEX__ 
   u8 data[length];
#endif
#ifdef __IMHEX__
    match (type) {
        (gltf_chunk_type_t::JSON): hex::type::Json<length> json;
        (gltf_chunk_type_t::BIN): u8 data[length];
    }      /**< The chunk data. */
 #endif
 ```
 You would need to also do all other parts that fail the unit tests.

File: patterns/gltf.hexpat [hunk 2]
```diff
@@ -64,7 +70,192 @@ fn gltf_format(gltf_chunk_type_t x)
...
+    ComponentType a12;
+    ComponentType a22;
+    ComponentType a32;
+    ComponentType a13;
+    ComponentType a23;
+    ComponentType a33;
+} [[static]];
+
+struct Mat4<ComponentType> {
+    ComponentType a11;
```
paxcut, line 125: consider doing something like
```cpp
struct Mat4<ComponenetType>{
    ComponentType a11,a21,a31,a41;
    ComponentType a12,a22,a32,a42;
    ComponentType a13,a23,a33,a43;
    ComponentType a14,a24,a34,a44;
    };
    ```
    or similar here and all the other ones to make the pattern more compact and easier to scroll and parse.
};

File: patterns/gltf.hexpat [hunk 3]
```diff
@@ -64,7 +70,192 @@ fn gltf_format(gltf_chunk_type_t x)
...
+        u64 accessor_offset = Json.accessors[accessor_index].byteOffset [[export]];
+    } else {
+        u64 accessor_offset = 0 [[export]];
+    }
+    view_offset = view_offset + accessor_offset;
+    u64 count_elements = Json.accessors[accessor_index].count;
+    ComponentType component_type = Json.accessors[accessor_index].componentType [[export]];
+
+    match (Json.accessors[accessor_index].type, component_type) {
+        ("SCALAR", ComponentType::BYTE): StrideType<Scalar<s8>, byte_stride> content[count_elements] @ view_offset + Offset;
```
paxcut, line 174: You want to avoid having to list all cases for each data type and each component type. There should be one case for SCALAR, one for VEC2, and so one with no sub-matches or nothing, just one line per type.  Instead of using the type asargument pass the enumeration so you can match the built in types at the lowest level possible. That way you only have to match them once instead of having to match them once for scalar, one for vec2 ,.... Note that you don't need to change the StrideType struct templates, only the ones for scalar, vec2,... 
Start by creating a new templated type for ComponentType (I would rename ComponentType enum to ComponentTypes.) that does the matching on the template argument. Then use the newly created type to populate scalar, vec2,vec3,...
I think the code will become cleaner and easier to read and work with.

paxcut: That would also not be optimal including matches for each type. Ypu can create only one match to test for content_type enum value
```cpp
struct ComponentType<auto component_type> {
    match (component_type) {
        (ComponentTypes::BYTE)                        :  s8 value;
        (ComponentTypes::UNSIGNED_BYTE)   :  u8 value;
        (ComponentTypes::SHORT)                      : s16 value;
        (ComponentTypes::UNSIGNED_SHORT) : u16 value;
        (ComponentTypes::UNSIGNED_INT)       :  u32 value ;
        (ComponentTypes::FLOAT)                       : float value;
    }
};
```
Then you can define Scalar and the rest like this
```cpp
struct Scalar<auto component_type> {
    ComponentType<component_type> x;
} [[static, sealed]];
```
with formaters to avoid the extra layers added. This creates the same data types but using only one match for the value types (int ,short,...) and one match for the data types (scalar, vector2,...).

paxcut:
> Can you maybe explain your last part of the comment a little bit?
> What do you mean with "access the binary data"? Access it to generate pattern data? Or access it to visualize it?

In order to visualize data, and by that i mean to use the visualize attribute on a pattern holding the data you must first generate pattern data. Right now the binary part of the file is converted into a pattern that is an array of bytes. Yes, it is a pattern, but no, it isn't what the format specifies as being stored there.

paxcut:
I have tested your pattern on files that are not the unit test file and found it very slow. It is true that the reason for it is mostly one of the changes I proposed. I wasn't considering the effect on evaluation time but moving the match for types to the last moment makes it have to check for matches at every value. To make up for my mistake I am attaching a patch to put the code back to how it would have been without that change so you can just copy paste it.

Even with the patch the code is still very slow for want it does. In one case it goes from 20 seconds to 2 minutes to display images, bufferviews and accessors. While it may be worth it for images, looking at buffers of bytes dont really say much about the data being stored. Some of the data are indices, some are vertices and so on which can be seen in the 3d visualizer model code I have added to the gltf pattern to visualize models stored.

[patch.txt](https://github.com/user-attachments/files/20960015/patch.txt)

WerWolv:
Thanks a lot! 
@paxcut Do you think we can merge this?

PR #413, opened 2025-06-28
Add support for medium and large mp4 files

klirktag:
The current implementation of mp4-pattern is very slow to deal with any larger mp4 files. Even loading a 50 megabyte mp4 video file is very slow and takes many minutes on a modern computer

The problem is the "mdat"  or MediaData box which is the biggest box contaning the raw media data have a very to slow implementation in the mp4-pattern language file.

This PR replaces the "mdat" box implementation with a much faster one, that so that we can work with files gigabytes in size instead of just 50 megabytes before everything gets way to slow

## Verify the speed improvement from this PR
Here is an example mp4 file of 270 megbyte
https://www.peach.themazzone.com/durian/movies/sintel-2048-surround.mp4
(found [here](https://durian.blender.org/download/))

Open it with the existing mp4-pattern. It take many minutes to open, if it manages to open at all.

Change to the implementation of mp4-pattern in this PR. The same file can now be opened in seconds.

### DIFF
--- a/patterns/mp4.hexpat
+++ b/patterns/mp4.hexpat
@@ -424,7 +424,7 @@ struct MovieBox : BaseBox {
 };
 
 struct MediaDataBox : BaseBox {
-    u8 data[while($ < endOffset)] [[sealed]];
+    std::mem::Bytes<boxSize - sizeof(size) - sizeof(type)> data;
 };
 
 struct Box {


paxcut:
Did you try using [std::mem::Bytes](https://docs.werwolv.net/pattern-language/libraries/std/mem.pat#std-mem-bytes)? it may be as fast or even faster.

klirktag:
> Did you try using [std::mem::Bytes](https://docs.werwolv.net/pattern-language/libraries/std/mem.pat#std-mem-bytes)? it may be as fast or even faster.

Not being an expert in ImHex, isnt Bytes and std::mem::Bytes the same? (I might be wrong)

paxcut:
Right on. Thank you for that. Is the size always going to be boxSize?

klirktag:
> Right on. Thank you for that. Is the size always going to be boxSize?

Yes, the size is always going to be boxSize. "mp4" is a continaer format, which means it contains other formats inside of it, but the formats inside of it can very

It typically contains h264, or h265, and aac, but can also include other codecs. They are stored in the "mdat" box. 

We cannot parse the inside of the mdat box, since it is dependent on the codec, which there are hundreds of different codecs, but we know for sure it is the boxSize that is the size of it

klirktag:
> No, i believe you, I just saw how it was using endOffset and made me wonder, but form the definitions at the top boxSize = endOffset-startOffset in all cases so it does work out. There are many places where endOffset is being used. They may not make much difference in the speed but if it all can be changed it will make things look more consistent. Just a thought, not really important or maybe not even possible.

That would certainly be interesting just for uniformity of the code base. Unfortunately I am not in a position to have time for such big refactoring at the moment, since I am working a full time job and also have small kids. The amount of time I have to contribute to open source is maybe 1-2 hours per month, late at night when the kids are sleeping :D

Maybe I can do such a refactoring in the future, in some years, when I have more time, or maybe someone else might have time to do that refactoring :)

klirktag:
> Don't worry about it, it isn't important as I mentioned, but what is important is the the changes made are correct. I looked at one of the example files linked which has a total size of 64657027 bytes and checked the values of boxSize and the size of the data array and they were not the same. BoxSize was 64312805 bytes and the data contains 64312797 of those bytes. An additional 8 bytes go to the size variable and the type string (mdat). It still is possible to use Bytes but the size should be `endOffset-$` or `boxSize - sizeof(size)-sizeof(type)`. I don't know which is correct but both cannot be.

I looked into this and realized this is correct. I choose the "data = boxSize - sizeof(size) - sizeof(type)" way, since it is easier to understand, and reduce the risk of people introducing bugs in the future

Thanks for helping me improve the PR so much! I just wanted a way to open bigger mp4 files, and now you made the code for it really good!

paxcut:
Looks good enough to merge. Thanks for your time.

PR #421, opened 2025-07-11
Add UEFI Firmare Volume Variable Store pattern

marcj303:
Add a pattern for UEFI Firmare Volume Variable store. This file type is commonly used with virtual machine UEFI variable files, like OVMF.fd used with QEMU. You could also extract a UEFI firmware binary from a flash device, search for the FV Variable Store, and set this pattern to the FV address.

### DIFF
--- a/patterns/uefi_fv_varstore.hexpat
+++ b/patterns/uefi_fv_varstore.hexpat
@@ -31,7 +31,7 @@
  */
 
 #pragma author Marc Jones
-#pragma UEFI Firmware Volume Variable Store
+#pragma description UEFI Firmware Volume Variable Store
 // #pragma debug
 
 import std.core;


patterns/uefi_fv_varstore.hexpat
```diff
@@ -0,0 +1,212 @@
...
+    // Align to the next 4-byte boundary for the next variable.
+    u8 pad[std::mem::align_to(4, sizeof(this)) - sizeof(this)];
+} [[name(this.Name), single_color]];
+
+
+// --- Main Pattern Entry Point ---
+
+EFI_FIRMWARE_VOLUME_HEADER FV_Header @ 0;
+
+if (std::core::formatted_value(FV_Header.FileSystemGuid) != "{FFF12B8D-7696-4C8B-A985-2747075B4F50}") {
```
marcj303, line 198: Is this the best way to check a GUID?

PR #425, opened 2025-07-31
Update FAS, WAS/WA3 pattern files and README.md

DimaLeon2000:
What I did is added the sound visualizer in the WAS/WA3 pattern, merged the frame bitmap list into the FAS struct, and fixed the typo in the ReadMe.


WerWolv:
Thank you!

PR #426, opened 2025-08-01
add embedded BMP and PNG parsing to ICO pattern

hexadecimalDinosaur:


WerWolv:
@hexadecimalDinosaur I believe you're the first one to use `import *` in the CI. I'll probably need to adjust the run parameters to include the pattern folder so they can be included. Adding something like `import bmp as BMP;` is planned but it's significantly more work than the simple `import *`. There's one workaround if you need it and that's the `IMPORTED` macro. That macro is only defined when a pattern is being imported by another one so you can `#ifdef` out parts of the pattern that you don't want to have run in that case. Not ideal of course but better than nothing

PR #428, opened 2025-08-02
Commodore BASIC

shewitt-au:
Commodore BASIC. 
PET, VIC20, C64, etc...

This is the first Pattern I've written. Checking for idiocy would be understandable.  That said, it works.


WerWolv:
Very cool! Is the hardcoded 65536 specific to the file format? Otherwise if there can be variable amount of data, using `while(!std::mem::eof())` for the array size might be better.

Could you also upload a test file and add your pattern to the readme please?

shewitt-au:
C64 only had 64k of memory. To be honest I didn;t think too much about it.
A line takes more than one byte, so it could be smaller. I will
implement your suggestions.

shewitt-au:
OK. I think I've done it. Sorry for the delay. I was worried about copyright stuff and had to hunt around for a progam I believe I can add.

WerWolv:
Thanks! The test file needs to be called `commodore_basic.hexpat.prg` for the CI to pick it up. Other than that it looks good for merging :)

shewitt-au:
OK. Done. I was wondering how the testing would work.

WerWolv:
That is great, yes :)

PR #435, opened 2025-08-25
Add a pattern for Roblox .pack files

mjarduk:
Essentially this adds a pattern for the `.pack` format legacy Roblox clients used for storing shader files and whatnot.

The file headers contain two fields that don't define any particularly useful metadata, so they're hidden for now.

[Reference files](https://archive.roblonium.com/Client/Windows/RobloxStudio/main%20%28roblox.com%29/2017/8.2017/August%2016%20%280.303.1.144437%29/version-0e0f1b65440542cf/shaders/)


WerWolv:
Thank you!

PR #437, opened 2025-09-04
Add terminfo pattern

nightowl286:
# Pattern

This is a pattern for the compiled *(legacy and extended)* term info entry files as defined by https://www.man7.org/linux/man-pages/man5/term.5.html.

These files are used on unix systems by a lot of different software *(most notably by [ncurses](https://invisible-island.net/ncurses/))* for determining the capabilities of your terminal emulator, and for knowing how to communicate with it.

Technically you can't count on the binary formats being compatible on different unix versions, however *(as I've been told)* those versions are pretty much dead nowadays so this format is common enough to work in like the majority of the cases.

## Checklist
- [x] A pattern for this format doesn't exist yet (or this PR improves the existing one).
- [x] The new pattern has been added to the relevant table in the Readme.
- [x] The new pattern has a description pragma (`#pragma description My pattern Description here`).
- [x] The pattern was associated with all relevant MIME types (using `#pragma MIME mime-type` in the source code).
  - Make sure to never use `application/octet-stream` here as that means "Unidentifiable binary data".
- [x] A test file for this pattern has been added to [/tests/patterns/test_data](/tests/patterns/test_data).
  - Try to keep this file below ~ 1 MB.

## Notes
- This format has two possible magic numbers *(at the same location in the file)* and each one slightly modifies a portion of the file *(not enough to warrant completely separate patterns)* but I don't know if the pattern language actually allows having multiple `#pragma magic` directives?
- The test files want you to use the `<pattern_name>.hexpat.<extension>` file name format, however terminfo files **do not** have a file extension, and I'm not sure if that might cause any problems?


WerWolv:
Very cool, thanks a lot!

For your questions:
- You can have as many `#pragma magic` and `#pragma MIME` as necessary. ImHex will try to match on each one of them
- Looks like right now it expects a file extension to work. I'd just append `.bin` and call it a day. Or anything else really that "fits"

WerWolv:
Thank you!

PR #441, opened 2025-09-13
Add TI calculator program pattern

Aversefun:


Aversefun:
Didn't see that someone else made one already :P

PR #442, opened 2025-09-16
DFIR Patterns

F01TECH:
Modified versions of /fs/ for semi-automated parsing of Disks/Volumes/Filesystems related to digital forensics and reporting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ Everything will immediately show up in ImHex's Content Store and gets bundled wi
 | DDS | `image/vnd-ms.dds` | [`patterns/dds.hexpat`](patterns/dds.hexpat) | DirectDraw Surface |
 | DEX | | [`patterns/dex.hexpat`](patterns/dex.hexpat) | Dalvik EXecutable Format |
 | DICOM | `application/dicom` | [`patterns/dicom.hexpat`](patterns/dicom.hexpat) | DICOM image format |
-| DISK_PARSER (DFIR) | `application/x-ima` | ['patterns/DFIR/DISK_PARSER.hexpat'](patterns/DFIR/DISK_PARSER.hexpat) | Recursive Disk/Volume/Filesystem parsing |
+| DISK_PARSER (DFIR) | `application/x-ima` | [`patterns/DFIR/DISK_PARSER.hexpat`](patterns/DFIR/DISK_PARSER.hexpat) | Recursive Disk/Volume/Filesystem parsing |
 | DMG | | [`patterns/dmg.hexpat`](patterns/dmg.hexpat) | Apple Disk Image Trailer (DMG) |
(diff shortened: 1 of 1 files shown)


F01TECH:
Great, thanks! 

Should we add 'DFIR_' to each, so there's not duplicates in the table?

F01TECH:
Good call.

In theory, the proposed patterns can replace the existing versions in /fs/.  When I was first developing, I had kept them separated due to temporary hard coded placements and never circled back around to consider full replacement. The entry point placements are all **0x0**, so will function as semi-automated or targeted (manual placement), just as the originals, but with all the enhancements.

Downside to full replacement:
  - The parsing of the file record entries...
    1. Adds considerable evaluation time on large volumes.  Under 5 seconds on 16GB or less, around 10-20 seconds on 100GB, and around 30-50 seconds on 500GB when capped at **4096 file record entries**.
    2. If someone were to use without checking the notes for each pattern, they may think they've parsed all entries or the pattern is incomplete/broken/inaccurate...
    3. Overkill for someone that simply wants to parse the MBR/GPT and any VBRs without diving deeper into the file system.    

I'll need to test and report back; Will setting the Root Dir/MFT MAX_NUM_ENTRIES to 0 break anything.

F01TECH:
I attempted to add a 'if DEEP_DIVE' for a quick fix, but too many errors to contend with right now. 

If someone doesn't want to parse FAT1/FAT2, RootDir, or MFT,  to speed things up, they can simply comment out the related calls or if the pattern offers a 'max number of' constant, set to zero.

For FAT32.hexpat:
  - // FAT_Entry FAT1[MAX_FAT_CHUNKS] @ FAT1_start_offset + 8;
  - // FAT_Entry FAT2[MAX_FAT_CHUNKS] @ FAT2_start_offset + 8;
  - // RootDirParser ROOT_DIR_ENTRIES[while(std::mem::read_unsigned($, 1) != 0x00)] @ root_dir_start;
  - // INFO_Overlay SFN_CLUSTER_LIST[MAX_SFN_CLUSTER_RELATIONS] @ FAT1_start_offset [[name("SFN <-> CLUSTER (DERIVED)")]];

For exFAT.hexpat:
  - const u64 MAX_FAT_CHUNKS = 0;
  - const u64 MAX_DIR_ENTRIES = 0;

for NTFS.hexpat:
  - // MFT_RECORD NEXT_RECORDS[while(fr_magic($) == NTFS_RECORD_TYPES::magic_FILE)] @ (nbs.mft_lcn * cluster_size + 12 * mft_rec_size) [[inline, static]];

That being said, its up to you whether or not we replace the simplicity/usability of the current /fs/ with the proposed patterns or keep it separate as /DFIR/... 

If we keep it separate, I can remove the duplicated file system patterns from the table and just have a reference to DISK_PARSER.hexpat?

PR #445, opened 2025-09-18
Credit to NE and improvements on NES

gmestanley:
I'm adding the improvements that I've been making to the NES pattern that I made. Support for NES 2.0 is fully added, among some other features like the Trainer that exists in some ROMs. Mapper names are not implemented yet.

I also added my credit to the NE pattern to not confuse people who take a look at it and see no credit.


WerWolv:
Thanks a lot! The NES pattern seems to fail the CI now due to it not finding a variable though

WerWolv:
No problem! Thanks for looking into it

gmestanley:
Pushed a fix.

PR #446, opened 2025-09-18
Add Python Pickle Pattern

ODeux2:
# Pattern

Pattern file for Python Binary Object Serialization Protocol aka Pickle.

The test file was generated from:
```python
import pickle

def main():
    # https://docs.python.org/3/library/pickle.html#what-can-be-pickled-and-unpickled
    data = {
        # built-in constants (None, True, False, Ellipsis, and NotImplemented);
        "none": None,
        "boolean_true": True,
        "boolean_false": False,
        "ellipsis": ...,
        "not_implemented": NotImplemented,
        # integers, floating-point numbers, complex numbers;
        "integer": 6942171,
        "float": 3.141592653589,
        "complex_number": 34+35j,
        # strings, bytes, bytearrays;
        "string": "Hello, World!",
        "unicode_string": "こんにちは-世界",
        "byte_string": b"byte string",
        "byte_array": bytearray(b"byte array"),
        # tuples, lists, sets, and dictionaries containing only picklable objects;
        "tuple": (1, 2, 3),
        "list": [1, 2, 3, 4, 5],
        "set": {1, 2, 3},
        "dictionary": {"key1": "value1", "key2": "value2"},
        # functions (built-in and user-defined) accessible from the top level of a
        #     module (using def, not lambda);
        "print_func": print,
        # classes accessible from the top level of a module;
        "range_class": range,
        # instances of such classes whose the result of calling __getstate__() is
        #     picklable (see section Pickling Class Instances for details).
        "range_class_instance": range(5, 10),
    }
    file_name = "pickle.hexpat.bin"
    with open(file_name, "wb") as file:
        pickle.dump(data, file, protocol = pickle.HIGHEST_PROTOCOL)
    print(f"Generated '{file_name}' test file.")

if __name__ == "__main__":
    main()
```

## Checklist
- [x] A test file for this pattern has been added to [/tests/patterns/test_data](/tests/patterns/test_data)


WerWolv:
Thank you very much!

PR #449, opened 2025-09-21
Add ESP32 image pattern

timschneeb:
# Pattern

Pattern file for ESP32 firmware images, including application segment headers.

Reference:
* https://github.com/espressif/esp-idf/blob/v5.5.1/components/bootloader_support/include/esp_app_format.h
* https://github.com/espressif/esp-idf/blob/v5.5.1/components/esp_app_format/include/esp_app_desc.h

## Checklist

* [x]  A pattern for this format doesn't exist yet (or this PR improves the existing one)
* [x]  The new pattern has been added to the relevant table in the Readme
* [x]  The new pattern has a description pragma (`#pragma description My pattern Description here`)
* [ ]  The pattern was associated with all relevant MIME types (using `#pragma MIME mime-type` in the source code)  
    * Make sure to never use `application/octet-stream` here as that means "Unidentifiable binary data"
* [x]  A test file for this pattern has been added to [/tests/patterns/test_data](/tests/patterns/test_data)    
    * Try to keep this file below ~ 1 MB


WerWolv:
Thanks a lot!

PR #451, opened 2025-09-26
Add support for KTX 1.0 textures

leximayfield:
Just what it says on the tin.  I've got a pattern for KTX 2.0 in progress, which is different enough to require its own pattern definition.  However, it's not quite ready yet, so this PR is just the 1.0 version.


WerWolv:
Thank you!

PR #456, opened 2025-10-13
[Pattern] add APE pattern

DexrnZacAttack:


WerWolv:
Thank you! Could you add the pattern to the readme table and add a test file if possible?

DexrnZacAttack:
> Thank you! Could you add the pattern to the readme table and add a test file if possible?

Sometime later yeah, sorry have been busy.

PR #462, opened 2025-11-08
PcapNG: Rewrite hexpat for bugfixing

antoniovazquezblanco:
I was debugging a piece of code that writes PcapNG files to disk. I noticed that the hexpat in ImHex was not correctly parsing custom blocks due to padding and size issues.

I started working on fixing this and realized the file was a bit too repetitive and was doing padding calculations in many places and coud be simplified. I got carried out and rewrote the file to use std::mem::AlignTo whenever needed and to recycle fields instead of replicating them in many places.

After some testing with my files I can confirm that my parsing issues were solved but I have not deeply tested Options parsing (my files do not have that section) although it should be ok.

I hope this is of your liking.
Thanks


WerWolv:
Thank you!

PR #464, opened 2025-11-22
UE5/UE4 Asset file parser.

Geramy:
Hey guys its been a while I know, here is our teams progress on the UE4/5 .uasset format parser.
There is a lot to it and its not as easy as saying well lets just do it. A big problem in .uasset for ue5 every C++ class has a small definition of the format its file is and if you dont have the class to parse that format you won't be able to parse it, but this includes most of the ue5 standard base classes. You will always be able to see an area that is defined for that class type. I also think it's possible to extrapolate the data still without the actually C++ class itself in hand but thats gonna take a lot more effort. As it is now this will parse any UE5 file and most of the stuff will be there. This tool helped me and my team a lot to figure out what's going on with the .uasset file format.


WerWolv:
Great work! Thank you very much!
If possible, could you add a test file to the respective folder and include your pattern in the readme table?

Roozi489:
Hey @Geramy, this pattern looks to be very helpful. But when I tried it, it fails for every uasset. Usually at parsing of the export properties (which is of course the most complex part). Sometimes it even prints this:
image
 
 What version of Unreal uassets should this be compatible with? Could you please share an uasset file this works correctly for?

Geramy:
Yeah thats not very kosher :P it was UE5.

PR #470, opened 2025-12-29
Create Agilent/Rigol pattern waveform parser

kenkit:
feat(patterns): add Agilent/Rigol wavebin waveform parser

Adds a new Pattern Language script to parse `.bin` waveform files exported from Agilent, Keysight, and Rigol oscilloscopes.

Features:
- Supports File Header versions 1, 3, and 10.
- Parses nested Waveform and Data headers.
- Automatically handles variable-length data buffers (float32 and u8).
- Includes auto-load pragmas for magic bytes 'AG' and 'RG'.
- Organizes sample data into inspectable arrays in the Pattern Data view.

Based on https://github.com/sam210723/wavebin/

### DIFF
--- a/patterns/wavebin.hexpat
+++ b/patterns/wavebin.hexpat
@@ -1,6 +1,8 @@
-#pragma pattern_name Agilent/Rigol Waveform
+#pragma author kenkit
+#pragma description Agilent/Rigol Waveform
 #pragma file_extension bin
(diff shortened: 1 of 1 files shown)


WerWolv:
Thanks a lot! Could you also upload a test file under `tests/patterns/test_data/wavebin.hexpat.bin` and add your pattern to the table in the readme?

Also fyi, the magic byte detection using `[ 41 47, 52 47 ]` doesn't work like that. You can however add two `#pragma magic` statements, one for each of the two sequences you want to match for

PR #475, opened 2026-01-09
Add pattern and magic for Arma 3 PAA

MrClock8163:
Adds pattern file for analysing the PAA texture format of the Arma 3 game.

A magic file is also included. So far it only has the PAA, but I plan on extending it with other Arma 3 formats.


WerWolv:
Thank you!

PR #476, opened 2026-01-11
feat: add SPC pattern

DexrnZacAttack:
I don't have an example file for this one, as Nintendo would not like that one bit... sorry.


WerWolv:
Completely fair, thanks a lot!

PR #477, opened 2026-01-11
More accurate variable declarations on nes.hexpat, add n64, gen, and gbx.hexpat

gmestanley:
Many variables in nes.hexpat were given more proper names and lengths (a more glaring example is "VT3xx" in ExtendedConsoleType; when I did this, general knowledge of the VT3xx line was still unclear, but now it's common knowledge that VT36x is the usual standard NES res and VT38x is high-res, and the latter does not have support in the format yet, so it's clearly VT36x). The constant FILLED_NES2_FLAGS was properly applied in all places that used it. The TxROM designation was more thoroughly elaborated with its more common types.<br>
DefaultExpansionDeviceType was added, primarily to highlight which ROMs use the device "multicart".

I have also added patterns for N64 and Mega Drive ROMs, as well as the fanmade GBX footer for GameBoy ROMs (similar to iNES/NES2.0 in the nes pattern).

### DIFF
--- a/patterns/nes.hexpat
+++ b/patterns/nes.hexpat
@@ -359,6 +359,7 @@ fn identifyMapper(u16 mapperValue, u8 submapperValue) {
 			(1): submapper = "No bus conflicts";
 			(2): submapper = "AND-type bus conflicts";
 		}
+	}
 	else if (mapperValue == 12) {
 		match (submapperValue) {
 			(0): submapper = "Supertone SL-5020B";


gmestanley:
nes.hexpat currently fails. This was shown in the last commit, but since in the first commit ttf.hexpat was failing instead (for some reason), I assumed the same thing was happening there. Will investigate later.

patterns/nes.hexpat
```diff
@@ -294,26 +318,47 @@ fn identifyMapper(u16 mapperValue, u8 submapperValue) {
...
+			else designation = "TxROM";
+		}
 	}
 	if (mapperValue == 3) {
 		match (submapperValue) {
 			(0): submapper = "Bus conflict";
 			(1): submapper = "No bus conflicts";
 			(2): submapper = "AND-type bus conflicts";
 		}
-	}
```
WerWolv: You accidentally removed a `}` here, that's why the unit tests are failing

gmestanley: Fixed

PR #500, opened 2026-02-21
Add msf/pdb file format

bwrsandman:
This is my best attempt at a pattern for .pdb file.

Tested with both an old MSF2.0 pdb and a very simple MSF7.0 pdb.


paxcut:
If you could please edit the Readme file to add your pattern like the others it would be greatly appreciated.

PR #519, opened 2026-04-13
fix: patterns were not handling the last byte of the input file correctly. (#518)

paxcut:
Changes are needed to pass pattern unit tests to latest changes in input file size handling.


dustinpower698-gif:
Now what is this everybody decide to quit talking in these chat rooms because I'm in it or what 
- [x] ```
`Jack potting ATMs 
`
```

PR #522, opened 2026-04-18
patterns: GVAS pattern for Unreal Engine 4+ games

scottanderson:
Unreal Engine 4+ Save Game file format.


paxcut:
Are the test files ok under the license implied? (gpl 2.0 i believe)

PR #528, opened 2026-06-01
[Pattern] Add Connectix Virtual Hard Disk (.vhd) format

XJDHDR:
I've created a pattern file for VHD files. I also added a basic pattern for MBR boot records.

There are three potential improvements I can think of for the VHD pattern:
- Have the pattern data for `modificationTime` display the time instead of an integer. I tried to get this working but couldn't figure it out. One issue is that the epoch for VHD's timestamps is the year 2000 rather than 1970.
- Add support for GPT boot records and partitions.
- Add complete support for Sparse (dynamically sized) and Differential images. Currently, only the footer is read, which is pretty much what a Fixed image consists of.

## Checklist
- [ ] The new pattern has been added to the relevant table in the Readme


paxcut:
the function used to print unix time is `type::impl::format_time_t()` so maybe you can try:
```cpp
fn TimeFormatter(std::time::EpochTime Epoch2000Time)
{
    std::time::EpochTime unixEpochTime = Epoch2000Time + 946684800;
     return type::impl::format_time_t(unixEpochTime );
//    return std::time::format(std::time::to_local(unixEpochTime), "%c");
//    return std::time::format(std::time::to_local(Epoch2000Time), "%c");
};
```

XJDHDR:
@paxcut 
> the function used to print unix time is `type::impl::format_time_t()`

Thanks for the help, but your suggestion didn't work. When I tried your code, the Pattern Data for `modificationTime` just said `Invalid`. After a bunch of debugging, it looks to me like there's something going wrong in the `builtin::std::time::to_utc` function. With this formatter function:
```
fn TimeFormatter(u32 Epoch2000Time)
{
    std::time::TimeConverter unixTime;
    unixTime.value = builtin::std::time::to_utc(Epoch2000Time + 946684800);
    return unixTime;
};
```

The Pattern Data window displays this:
`union std::time::TimeConverter { time = struct std::time::Time { sec = 0, min = 0, hour = 0, mday = 0, mon = 0, year = 0, wday = 127, yday = 1280, isdst = true }, value = 2343096923887082024718 }`

paxcut:
I tried this simple test:
```cpp
std::time::EpochTime unixEpochTime = 946684800;
std::print("{}",type::impl::format_time_t(unixEpochTime));
```
and it printed `Sat Jan  1 00:00:00 2000`. A million more seconds prints `Wed Jan 12 13:46:40 2000`  which seems correct (`12x24x60x60 = 1,036,800`). Make sure your epoch2000 time in seconds makes sense.
Watch out for:
1) std::time::EpochTime is defined as u32 which may be overflowing if it needs a u64 value.
2) endianness of the epoch2000 value read. print it to make sure it agrees with data read.

XJDHDR:
I figured out what is going on, and it might be a bug in either the `builtin::std::time::to_utc` function or the `TimeConverter` union.

Because VHD's data is in Big Endian format, I have `#pragma endian big` defined in the pattern. I've just found that one of those appears to be designed to assume it is Little Endian. 

I tried taking the pattern and test VHD from the PR, then uncommented the modificationTime `format` attribute and modified it's function to:
```
fn TimeFormatter(std::time::EpochTime Epoch2000Time)
{
    std::print("{}",type::impl::format_time_t(946684800));
};
```

With that pragma set to Little, `Sat Jan  1 00:00:00 2000` is printed to console. Setting that pragma back to Big Endian causes `Invalid` to be printed instead.

I then changed the function to:
```
fn TimeFormatter(std::time::EpochTime Epoch2000Time)
{
    std::time::TimeConverter unix2000Time;
    unix2000Time.value = builtin::std::time::to_utc(946684800);
    
    std::print("{}", unix2000Time);
};
```
Changing Endianness prints the following:
Little: `union std::time::TimeConverter { time = struct std::time::Time { sec = 0, min = 0, hour = 0, mday = 1, mon = 0, year = 100, wday = 6, yday = 0, isdst = false }, value = 432455515407122432 }`
Big _: `union std::time::TimeConverter { time = struct std::time::Time { sec = 0, min = 0, hour = 0, mday = 0, mon = 0, year = 0, wday = 0, yday = 1536, isdst = true }, value = 432455515407122432 }`

I'm not sure however if there's something I'm doing wrong, or if it's a bug that needs to be fixed.

paxcut:
The use of little endian in the time functions goes beyond the items you mentioned and extends even to the builtin functions as evidenced by the fact that using big endian when calling the builtin format function causes invalid to be printed. Therefore it is not clear there is a bug in the items mentioned, you just need to set the byte order temporarily when calling the time functions like this:

```cpp
fn TimeFormatter(std::time::EpochTime Epoch2000Time)
{
    std::core::set_endian(std::mem::Endian::Little);
    std::print("{}",type::impl::format_time_t(946684800));
    std::core::set_endian(std::mem::Endian::Big);
};
```
which should work when big endian is set globally if both std.core and std.mem are imported. The second example appears to be working when endian is set to little, hours, minutes and seconds are set correctly to zero. month day is also correctly set to 1, month is zero, week day is 6 and year day is zero. The year being 100 I don't really understand cause I would have expected 32 but seeing how all other values are correct I suspect that I just don't know enough about these time calculations.

XJDHDR:
I did some more tests and found that the builtin functions handle endianess perfectly fine. The problem happens in `std::time::to_utc` when the value returned by the `builtin::std::time::to_utc` function is put into the `TimeConverter` union. I'm assuming this union is expecting the u128 to be in Little Endian format. 

Otherwise, thanks for the help. I've managed to write the time formatter function so that it correctly converts the integer into a time string. I'll see if I can add some sort of Dynamic and Differential Image support before I add another commit.

paxcut:
The `std::time::Time` structure is written with values increasing in significance from lowest to biggest which means that the converter union will only work when the u128 value has least significant values first so it must be little endian. That poses no problem since you have complete control of what order to use when and where.

I am glad you found a way to get it to work. I may not always be of much help, but when I can, I do it happily. GL!

WerWolv:
Thank you!

PR #529, opened 2026-06-03
Add PSP Self pattern

LiEnby:
this is the (signed) executable for the PlayStation Portable game console;
 .. part of the "data.psp" section in certain EBOOT.PBP files;


WerWolv:
Thank you! Could you add a `#pragma description` and `#pragma author` tag to your pattern please for our UI?
Also please add your pattern to the table in the readme and (if possible at all) upload a test file to the `tests/patterns/test_data` folder.

The test file is not required if these files can't be shared but it helps making sure our patterns don't break :)

LiEnby:
i dont know of any psp selfs i could add that wouldn't be kinda copyrighted to hell ..

paxcut:
As mentioned by @WerWolv , the input test file would be nice but it is not necessary. There are many patterns that do not include a test file for the same reason.

We do however ask that you add your pattern to the README file so that the pattern gets listed. Look at some other pattern to fill in the required fields . 

Last, there is some pattern failing unit tests but it seems unrelated to yours and may just need to have the branch updated which is something I can do.

Nemoumbra:
We have a dedicated `type::Magic` thingie, by the way. Do you think it'd be better to have something like `type::Magic<"~PSP"> magic;`?

I asked @GrayJack about the sample files. Apparently `https://github.com/PSP-Arkfive/ARK-5/releases/tag/latest` can work.
> Inside the zip, there is a folder called flash0
There are all ARK CFW modules
Almost sure that all of them are in ~PSP format

Nemoumbra:
I'll also just leave it here: https://gist.github.com/Linblow/8639cb76ad6f813dc56d15cd40cf48ab

LiEnby:
are the flash0 contents in ARK-5 not simply modified versions of the SCE modules (?)

> We have a dedicated `type::Magic` thingie, by the way. Do you think it'd be better to have something like `type::Magic<"~PSP"> magic;`?
> 
> I asked @GrayJack about the sample files. Apparently `https://github.com/PSP-Arkfive/ARK-5/releases/tag/latest` can work.
> 
> > Inside the zip, there is a folder called flash0
> > There are all ARK CFW modules
> > Almost sure that all of them are in ~PSP format

are the ARK-5 flash0 not simply modified versions of the SCE modules? i thought they were ...

GrayJack:
> are the ARK-5 flash0 not simply modified versions of the SCE modules? i thought they were

No, this folder only contains CFW modules, ARK always modifies SCE modules on runtime, during module load

LiEnby:
i would take a look, however it seems the github releases page specifically does not seem to be loading properly for me at the moment, not sure exactly why, everything else works, so perhaps later.

WerWolv:
Thank you!

PR #531, opened 2026-06-18
Add CGX Shader File

LiEnby:
Used in playstation mobile games


paxcut:
If possible it would be great if you can provide a small test input file for the pattern but it is not critical. We do ask that you edit the Readme file to include your entry like the other entries do.

LiEnby:
where would i put the example CGX in here?

paxcut:
test files usually go in the `tests/patterns/test_data` folder using the name convention seen there like e.g. `bmp.hexpat.bmp`

LiEnby:
okay done,

paxcut:
As you can see your pattern is failing unit tests. If you look at the run log you will see

  1/284 Test   #3: Patterns/CGX.hexpat ........................................................***Failed    0.07 sec
Parsing pattern CGX.hexpat without executing it
No description pragma found in pattern file
Please add a #pragma description <your description> tag to the pattern!

It appears  that the pragma description is mandatory requirement to be part of t\e pattern library.

LiEnby:
added, also updated some feilds to be more up to date w what i found later in SnowPME,

WerWolv:
Thank you!

PR #532, opened 2026-06-20
Add CSNAP pattern

FireBlade211:
Adds the pattern for the clipboard snapshot format from my new app [ClipboardEdit](https://github.com/FireBlade211/ClipboardEdit/tree/main/README.md#snapshot-format)


paxcut:
Hello, I looked at the pattern and downloaded you app. I have some questions and suggestions that in my opinion will make it easier to maintain and expand the pattern to include more types of data. 

For that purpose you can swap the function to convert ids to formats with an enumeration. The enumeration format automatically defaults to the enumeration names which makes the translation both practical and faster except for two differences with your code. The default format includes the enumeration name as e.g. `ID::CF_TEXT`. The second difference is that for values not in the enum the default prints `ID::???`. Both of them can be fixed easily. 

First the enum is much simpler and easier to add new entries.
```cpp
enum ID : u32 {
    CF_TEXT = 1,
    CF_BITMAP,
    CF_METAFILEPICT,
    CF_SYLK,
    CF_DIF,
    CF_TIFF,
    CF_OEMTEXT,
    CF_DIB,
    CF_PALETTE,
    CF_PENDATA,
    CF_RIFF,
    CF_WAVE,
    CF_UNICODETEXT,
    CF_ENHMETAFILE,
    CF_HDROP,
    CF_LOCALE,
    CF_DIBV5,
    CF_OWNERDISPLAY = 0x80,
    CF_DSPTEXT,
    CF_DSPBITMAP,
    CF_DSPMETAFILEPICT,
    CF_DSPENHMETAFILE = 0x8E,
};
```
The rest of the code is the same except for the format function.
```cpp
fn id_to_format(auto v) {
    ID id = v;
    str result = std::format("{}",id);
    u32 resultLength = std::string::length(result)-4;
    result = std::string::substr(result,4,resultLength);
    if (result == "???")
        return std::format("{:#X} {}",v,v);
    return result;
};
```
The function converts the u32 value to an enum then converts that to a string that has the extra stuff trimmed. Finally the function formats values that end up as question marks.

Again these are only suggestions that you don't need to accept if you don't want to.

The questions are two, First about the reason why you are calculating all the time formats from scratch when the pattern language provides types and functions to do it for you and do it faster in c++ compiled code. The following three lines are equivalent to all the date and time code you added and it results in better looking format.

```cpp
 type::FILETIME filetimeTakenUtc[[format("to_utc")]];


fn to_utc(auto v) {
    type::time64_t unix_time = std::time::filetime_to_unix(v);
    return type::impl::format_time_t(unix_time);
};
```
This prints the date/time as `Sun Jun 21 10:11:39 2026` instead of `2026-06-21 10:11:39`. 

 If you decide you want to use these suggestions you will need some imports as well:
```cpp
import std.io;
import std.core;
import std.time;
import type.time;
import std.string;
```

The final question is about the file uploaded as a test input for the pattern. I had to download your app to create a test file because the file uploaded seems to be a copy of the pattern instead of a valid csnap file. Your app ran very well and it looks like it would be useful and it is well done.

If you could please upload a valid test file then the pattern can be added to the library right away. Thanks!

FireBlade211:
1. I used an enum originally, but after realizing that ImHex would display it something like, `CLIPBOARD_FORMAT::CF_TEXT`, I switched to a function to make it cleaner.
2. I didn't know the pattern language included a built-in type for a FILETIME, but thanks for the suggestion.
3. Whoops, looks like I misread that contributing guide. I've fixed it now with a basic test file.

paxcut:
> I used an enum originally, but after realizing that ImHex would display it something like, CLIPBOARD_FORMAT::CF_TEXT, I switched to a function to make it cleaner.

yes, i mention that in the write-up and showed how to resolve it. Perhaps you missed that part? It is not important, but wanted to make that clear. I'll merge it as soon as the tests pass. Thanks again.

paxcut:
Your pattern is failing unit tests. Based on the error message it looks like it is missing a required pragma description that all patterns in the library must have.

```
  47/284 Test  #48: Patterns/csnap.hexpat/csnap.hexpat.csnap ...................................***Failed    0.47 sec
Executing pattern csnap.hexpat using test file csnap.hexpat.csnap
    [INFO]  Pattern exited with code: 0
No description pragma found in pattern file
Please add a #pragma description <your description> tag to the pattern!
```

FireBlade211:
Ok, now the formats use an enum, and I also added the required #pragmas.

PR #534, opened 2026-06-24
patterns/elf: Fix off by 1 error in shdr payload size checking

AlexGuo1998:
The old check `sh_offset + sh_size < std::mem::size()` fails when the payload ends just before EOF. This can happen for hand-crafted files when `shdr` is not at the end.

### DIFF
--- a/patterns/elf.hexpat
+++ b/patterns/elf.hexpat
@@ -676,7 +676,7 @@ struct Elf32_Shdr {
     Elf32_Word sh_addralign;
     Elf32_Word sh_entsize;
 
-    if (sh_size > 0 && sh_offset + sh_size < std::mem::size()) {
+    if (sh_size > 0 && sh_offset + sh_size <= std::mem::size()) {
     	if (sh_type == SHT::NOBITS || sh_type == SHT::NULL) {
 		// Section has no data
 		} else if (sh_type == SHT::STRTAB) {
@@ -751,7 +751,7 @@ struct Elf64_Shdr {
     Elf64_Xword sh_addralign;
     Elf64_Xword sh_entsize;
 
-    if (sh_size > 0 && sh_offset + sh_size < std::mem::size()) {
+    if (sh_size > 0 && sh_offset + sh_size <= std::mem::size()) {
     	if (sh_type == SHT::NOBITS || sh_type == SHT::NULL) {
 			// Section has no data
 		} else if (sh_type == SHT::STRTAB) {


paxcut:
I guess what counts here is finding a test input file that is a valid elf executable that needs this patch for the pattern to run correctly because I am having trouble following the arguments made.

Where can I find such  file? I assume you have one, can you share it? If not, can you describe how to make one?

Setting aside the elf pattern or any pattern, strictly speaking, the last address of any file is `std::mem::size()-1` so that seems to imply that <= would mean that the file was overflown. so that if `sh_offset + sh_size == std::mem::size()` then `u8 data[sh_size] @ sh_offset;` should be generating an error of reading past the eof.

AlexGuo1998:
> Setting aside the elf pattern or any pattern, strictly speaking, the last address of any file is `std::mem::size()-1` so that seems to imply that <= would mean that the file was overflown. so that if `sh_offset + sh_size == std::mem::size()` then `u8 data[sh_size] @ sh_offset;` should be generating an error of reading past the eof.

I believe that's not the case here. Say we have a file of 4 bytes, `std::mem::size()` returns 4. `sh_offset = sh_size = 2` (`2 + 2 <= 4`), which places a 2-byte pattern at offset 2, and is totally valid.

> I guess what counts here is finding a test input file that is a valid elf executable that needs this patch for the pattern to run correctly because I am having trouble following the arguments made.
> 
> Where can I find such file? I assume you have one, can you share it? If not, can you describe how to make one?

I'm unable to share the original file I encountered, but generating one is easy. The idea is to build a normal ELF, and duplicate, say, the payload of `.shsrtab` to the end, and update `sh_offset` to point to the new bytes.

Screenshot From 2026-06-24 17-03-32

[test.zip](https://github.com/user-attachments/files/29285731/test.zip)

BTW, the comparison with `std::mem::size()` in `phdr` is also using `<=`

https://github.com/WerWolv/ImHex-Patterns/blob/537f7abbf6c610eccfa5119ffbdc80ee9f25fecf/patterns/elf.hexpat#L599

... and that was fixed previously just in ecd34d35b22aa142650c97eaeb93f8c6c1b679db

paxcut:
ok yes, that's correct and i was mistaken. offset + size is past the end of the data and the last valid data address is offset + size -1. comparing  sh_offset + sh_size with std::mem::size() is comparing two sizes, it is just that the first  size being compared needs to have its offset taken into consideration to make it comparable to the second size, but the condition is that the first size must be smaller than or equal to the second size.

The reason this wasn't detected before  is the fact that sh_offset and sh_size are not the offset and size of shdr. That can be seen here
```cpp
Elf64_Shdr shdr[ehdr.e_shnum] @ ehdr.e_shoff;
```
sh_offset and sh_size are the offsets and size of some data associated with shdr that is normally located way before the start of shdr which is why it uses a pointer to define it. In most cases the first size is not only smaller but much smaller than the second so no error exists.
Only under some specific conditions that associated data will be located at the very end of the file and that's the only place where the incorrect size comparison will create problems and O am guessing those are the types of elf execs you found.

PR #539, opened 2026-07-15
Add formatters to GGUF pattern

liberlanco:
Add formatters to make it more useful during research and to avoid constant expanding of items. Formatters for metadata KV pairs, and formatters for tensor blocks with offsets.

Also increase patters/array limit to fit some GGUF files.

Screenshots after the fix on test data:
screenshot metadata

screenshot token info


WerWolv:
Thank you!

PR #543, opened 2026-07-22
added GTA: San Andreas save game block pattern

S1ngularity96:
Added GTA: San Andreas save game pattern.


WerWolv:
Thank you!
Could you upload a test file to the respective folder that we can run your pattern on for tests?

WerWolv:
Thank you!

PR #551, opened 2026-08-02
disassemblers: Add Zelog Z80 disassembler

smiRaphi:
Adds a Zelog Z80 disassembler.
Note that I haven't implemented the IX & IY prefixes as that would more than quadruple the opcode count.
I've also added all undefined instructions I could find that aren't just `NOP`s, including duplicate `NEG`s, `RETN`s, `RETI`s and the fixed value `IM`s, `IN` & `OUT`

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,3 +376,4 @@ Everything will immediately show up in ImHex's Content Store and gets bundled wi
 | Name | Path | Description |
 |------|------|-------------|
 | 8051 | [`disassemblers/8051.json`](disassemblers/8051.json) | Intel 8051 Architecture |
+| Z80 | [`disassemblers/z80.json`](disassemblers/z80.json) | Zelog Z80 Architecture |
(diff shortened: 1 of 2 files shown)


WerWolv:
Thanks a lot! Really cool seeing this come together!

For the prefixes, there's actually some built-in support for them:
https://github.com/WerWolv/ImHex-Patterns/blob/master/disassemblers/jvm.json#L1038

Basically, for each opcode, it will first try to parse one or more of the prefixes and then try the opcodes. Prefix mnemonics are then simply prepended to the opcode that comes after them so you'd only need to define these two prefixes instead of duplicating the opcodes 2^N times

smiRaphi:
I see but, as I understand it, the IX/IY prefix is implicitly generated by the assembler so just prefixing it wouldn't be valid but I guess still better than having it as it's own instruction

Edit: also it's even more complicated than that since the prefix can have extra data after the actual opcode:
https://en.wikipedia.org/wiki/Z80_instruction_set#Encoding_order

ABratovic:
@smiRaphi FYI, it's a Zilog Z80, not Zelog :smiley:

This URL should help with the unknown opcodes
https://clrhome.org/table/

PR #552, opened 2026-08-02
disassemblers: add lua 5.0 through 5.5

neptuwunium:
Generated from opcodes listed in lopcodes.h and lopcodes.c using [this script](https://codeberg.org/neptuwunium/scripts/src/branch/develop/GenerateLuaOpcodes.py).

There is still some improvements possible with formatting, I couldn't get it to properly display values (`0b0100'0001` would display as `loadk 0x40` when it's `loadk 0x01`)

Left in the opcode comments from lua as metadata for if anyone wants to go through and make formatters.

image

### DIFF
--- a/disassemblers/lua50.json
+++ b/disassemblers/lua50.json
@@ -1,0 +1,322 @@
+{
+  "name": "Lua5.1",
+  "includes": [],
+  "prefixes": [],
+  "opcodes": [
+    {
+      "mnemonic": "move",
+      "format": "",
+      "mask": "AA00'0000 BBAA'AAAA CBBB'BBBB CCCC'CCCC",
+      "metadata": {
+        "comment": "A B R(A) := R(B)",
+        "opmode": "iABC"
+      }
+    },
+    {
+      "mnemonic": "loadk",
+      "format": "",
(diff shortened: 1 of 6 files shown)


WerWolv:
Thank you!

How did you try to do the formatting? As I see right now, you're doing any of the parameter formatting at all.
The bit placeholders you use in the `mask` field all extract the bits marked bits from the data and then those bits get cat'ed together to form the number the corresponding variable gets set to in the end
(e.g, `1100'0011` with mask `AABBBBAA` results in `A  = 0b1111` and `B = 0b0000`)

neptuwunium:
> How did you try to do the formatting?

`iABC` (which loadk is) would be `0x{A:02X}, 0x{B:02X}, 0x{C:02X}`, with the same mask that is present here.

iABC in 5.4 and later is defined as

```
iABC          C(8) | B(8) | k | A(8) | Op(7) |
```

and prior to 5.4 as

```
iABC          C(9) | B(9) | A(8) | Op(6) |
```

judging from the explanation i think i need to bitshift the whole instruction as an unsigned 32-bit value since that's what it's technically processed as.

WerWolv:
Converting this to a draft until the argument formatting has been added

PR #553, opened 2026-08-04
Half-life .dem format pattern

den4iklovelinux:
Simplified demo data pattern. I think this will be nice for those people who wants to write their own parser (like i do, so i wrote this hexpat)
In the future I plan to write patterns for other engine formats, like saves, etc.


WerWolv:
Thank you! Looks good

Could you add the pattern to the table in the readme and also, if possible, upload a test file to /tests/patterns/test_data so we can make sure the pattern doesn't break?

WerWolv:
Thanks!

den4iklovelinux:
i think ill upgrade this pattern in the future, not now cus im not get how to work with the netmessages.

WerWolv:
Feel free to open another PR later to improve your pattern!

PR #554, opened 2026-08-06
Fix: change gb.hexpat ramSize type from u16 to u32

cocoastardust:
Changed the type for `ramSize` from `u16` to `u32` to fix overflow for larger RAM sizes in gb.hexpat.


WerWolv:
Thank you!

PR #555, opened 2026-08-06
Updated goldsrc .dem pattern

den4iklovelinux:
Whats changed:

- 0 and 1 type frames data was extended for refparams, usercmd and refparams data.
- More types of structs was added (for almost every frame types) like clientdata, sounddata and etc
- Names refactoring
- Camprefs was extended for more params (forward, right and up)

### DIFF
--- a/patterns/hldem.hexpat
+++ b/patterns/hldem.hexpat
@@ -9,7 +9,7 @@
 #pragma description GoldSRC DEM
 #pragma magic [ 48 4C 44 45 4D 4F ] @ 0x00
 #pragma endian little
-#include <std/io.pat>
+
 
 struct header_t
 {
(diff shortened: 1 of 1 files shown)


den4iklovelinux:
AFAIK usermsg parsing will be too complicated, this pattern already takes a long time to execute (like 6 seconds on 10mb file), i think with parsing messages this will be much longer.

den4iklovelinux:
> much longer 


i mean even in my C/++ code this isnt so fast, in 108mb demo parsing ONE type of usermsg (PlayerDeath_Msg) takes like 1 second to read all delta-compressed strings of this demo. I shudder to think how much longer it would take using patterns, and how much the limits would have to be raised to accommodate that.

WerWolv:
The Pattern Language is not extremely performant generally speaking. So it taking a few seconds to parse a decently sized file is pretty common, especially if it's generating a lot of patterns.

One thing you can do is add it as a parsing option. Define a bool `in` variable that you can enable from the UI in ImHex. And only if it's enabled, do the extended parsing

den4iklovelinux:
> The Pattern Language is not extremely performant generally speaking. So it taking a few seconds to parse a decently sized file is pretty common, especially if it's generating a lot of patterns.
> 
> One thing you can do is add it as a parsing option. Define a bool `in` variable that you can enable from the UI in ImHex. And only if it's enabled, do the extended parsing

thanks, ill think about it! I tried to parse large 40min 108mb demo, but pattern hit the limit.... When i tried to copy error ImHex just crashed

WerWolv:
Sorry about that, if you can send the error log my way I can take a look

den4iklovelinux:
> Sorry about that, if you can send the error log my way I can take a look

I think its more like valve demos format and my pattern fault, anyway, when i tried to reproduce this bug imhex aint even crashed lol

WerWolv:
Even if it's your pattern causing it, ImHex should never be crashing :P
Could be though that your ran out of memory if there were too many patterns created

PR #557, opened 2026-08-10
patterns: Add Rockchip Vendor Storage

trzen:
Adds the Rockchip Vendor Storage format. This is a format designed to store vendor data like serial numbers and MAC addresses.

Adds a test file aimed at a eMMC/NAND configuration containing 2 MAC addresses (`AA:BB:CC:DD:EE:FF`) and a serial number (`0123456789`).

The pattern file contains some comments regarding `type::Magic`, happy to remove those if needed.


WerWolv:
Thank you!

PR #565, opened 2026-08-17
patterns/flp: Fixed bugs and added plugin stuff for native and VST stuff

PentaChess413:
-Fixed the fact that only the first plugin was parsed
-Added a bunch of other stuff
-Somewhat proper UTF-16 LE handling, might still break. Mail me any query.


WerWolv:
Thank you!

PR #566, opened 2026-08-18
add param.sfo

LiEnby:
- PSP/PS4/PS3/PSVita SFO file
used for system information and metadata for applications

- example file is from my own vita homebrew software .. ([GcToolKit](https://github.com/oestriot/GcToolKit))


WerWolv:
Thank you!

PR #570, opened 2026-08-20
patterns: Add OpenSSH private key support

LucasChollet:
This pattern file only supports keys without a passphrase and I only tested it with ed25519 keys. It might not work with every other key types.

Thanks @neptuwunium for helping out!


WerWolv:
Thank you!

PR #572, opened 2026-08-24
Add pattern for Dig or Die uncompressed save files

ddmitv:
Adds a pattern for viewing [Dig or Die](https://store.steampowered.com/app/315460/Dig_or_Die) uncompressed save files. The format is reverse engineered from the game's decompiled C# code and a bit of C++ `.dll` decompilation. The game uses the C# `BinaryFormatter` to store some parts of the game state, so I have made a workaround for this by just skipping the `BinaryFormatter` metadata and only parsing useful data.

The original save files are compressed using the LZF algorithm, so this pattern requires the uncompressed save files.
Because uncompressed saves are quite large (~19 MB), I have not included a test file for this pattern.

If you are worried about copyright concerns, the game developer has explicitly stated that community reverse engineering is "ok as long as it's not official, and it does not help pirating the game (playing it without buying it)".

I'm also considering next submitting a PR to the main ImHex repo to add a decompression function for the LZF algorithm (since the algorithm is simple, no extra dependencies), but I understand that this may be too niche of a feature and might possibly result in only me using it. Let me know if it's not a good fit for the core language. I completely understand that and won't submit it!

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ Everything will immediately show up in ImHex's Content Store and gets bundled wi
 | DEX | | [`patterns/dex.hexpat`](patterns/dex.hexpat) | Dalvik EXecutable Format |
 | Devil May Cry HD Collection | | [`patterns/Devil May Cry HD Collection`](patterns/Devil May Cry HD Collection) | 3D Model files used in Devil May Cry 3 HD Collection |
 | DICOM | `application/dicom` | [`patterns/dicom.hexpat`](patterns/dicom.hexpat) | DICOM image format |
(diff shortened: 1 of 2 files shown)


WerWolv:
Thank you! 

Not having a test file here is alright if the size can't be reduced. As for adding LZF support, it looks to me like it's a pretty common (although somewhat old) algorithm. If implementing it doesn't require a large amount of code being added, I don't see a reason to not have it

ddmitv:
Thanks. Once the LZF decompression function is added to the core, I think I can add support for compressed save files and with it, add a test file. Although, regular compressed saves are still around ~2 MB, but I think I can craft an optimized for compression ratio save file to fit into the requirement.

PR #574, opened 2026-08-25
Update Dig or Die save pattern to support compressed save files + add test file for it

ddmitv:
Adds support for LZF-compressed save files using the `hex::dec::lzf_decompress` function to decompress data into a section and then parse the actual save data from it.

Also adds a test file for this pattern (~98 KB). Since `hex::dec::lzf_decompress` is not available in the unit test environment, I have crafted an uncompressed save file with only a 64x64 world grid (instead of 1024x1024) so it could actually fit within 100 KB. Used the `__PL_UNIT_TESTS__` preprocessor variable to set the world width and height `in` variables inside the unit test to match the corresponding test file's world parameters (the save file doesn't contain world dimensions).


ddmitv:
@WerWolv, can you please check this PR?

WerWolv:
Thank you!

PR #575, opened 2026-08-28
Add FLP file pattern to README

PentaChess413:
Allows my FLP pattern to be downloaded from the Content Store.


WerWolv:
Thank you!
The table doesn't affect the content store though so if the file isn't showing up, something else is wrong. I'll take a look at the backend

PR #576, opened 2026-08-30
patterns: Add Kerberos5 keytab pattern

Mikle-Bond:
Couldn't find a keytab parser, so I made one. It is based on description from https://www.ioplex.com/utilities/keytab.txt.
It follows the Version 2 of the format. Version 1 uses native host byte order, which I'm not sure how to represent in a hexpat file.
Also tried to format the displayed values to match the output of `klist -e -t -k tests/patterns/test_data/keytab.hexpat.keytab`.


WerWolv:
Thank you!

Native byte order would require you to somehow detect what byte order is used or asking the user for it using an `in` variable and then calling `std::core::set_endian(std::mem::Endian::big)` or `std::core::set_endian(std::mem::Endian::Little)` instead of using `be` for each member.


Your pattern seems to fail the unit tests right now btw, could be related to the `return` you're using, I think you rather want `break`. Here's the log:
```
[5](https://github.com/WerWolv/ImHex-Patterns/actions/runs/33297149678/job/99249978838?pr=576#step:6:256)
Executing pattern keytab.hexpat using test file keytab.hexpat.keytab
    [ERROR] [ Stack Trace ]
    [ERROR] <Source Code>:59:22 at address 0x008E
    [ERROR] 59 |     char value[length];
                          ^
    [ERROR] <Source Code>:57:8 at address 0x008C
    [ERROR] 57 | struct CountedString {
            ^^^^^^^^^^^^^
    [ERROR] <Source Code>:78:5 at address 0x008C
    [ERROR] 78 |     CountedString realm;
         ^^^^^^^^^^^^^
    [ERROR] <Source Code>:76:8 at address 0x008A
    [ERROR] 76 | struct Principal {
            ^^^^^^^^^
    [ERROR] <Source Code>:98:5 at address 0x008A
    [ERROR] 98 |     Principal principal;
         ^^^^^^^^^
    [ERROR] <Source Code>:96:8 at address 0x008A
    [ERROR] 96 | struct EntryContents {
            ^^^^^^^^^^^^^
    [ERROR] <Source Code>:127:9 at address 0x008A
    [ERROR] 127 |         EntryContents entry [[fixed_siz
              ^^^^^^^^^^^^^
    [ERROR] <Source Code>:128:5 at address 0x008A
    [ERROR] 128 |     }
          ^
    [ERROR] <Source Code>:116:8 at address 0x0086
    [ERROR] 116 | struct Entry {
             ^^^^^
    [ERROR] <Source Code>:146:5 at address 0x0086
    [ERROR] 146 |     Entry entries[while(!std::mem::
          ^^^^^
    [ERROR] <Source Code>:140:8 at address 0x0000
    [ERROR] 140 | struct Keytab52 {
             ^^^^^^^^
    [ERROR] <Source Code>:149:1 at address 0x0000
    [ERROR] 149 | Keytab52 keytab @ 0x0;
      ^^^^^^^^
    [ERROR] Error happened with cursor at address 0x008E
    [ERROR] 
    [ERROR] runtime error: Array expanded past end of the data.
  -->   in <Source Code>:59:22
59 |     char value[length];
                          ^
Non-zero exit code returned from execution. (Exit Code: 1)!
59:22  runtime error: Array expanded past end of the data.
  -->   in <Source Code>:59:22
59 |     char value[length];
                          ^
```

Mikle-Bond:
woah  ._.

So, after digging a bit, I think I figured out where the problem is. It is in the line `EntryContents entry [[fixed_size(entry_size)]]; $ += entry_size;`. 

Looks like `patterns_test` executable and the ImHex GUI treat the `fixed_size` attribute somehow differently. When I was making the pattern it kept failing at wrong offsets, and I figured that members marked with `fixed_size` do not advance the offset. I bump it manually, and pattern works fine in ImHex. I ran `patterns_test` with strace, and looking at what kind of data it looks at, the offsets become wrong again.

Now that I look at it, the only other pattern file that also uses `fixed_size` is for sqlite, and there it it written as `[[fixed_size(le u16(globalPageSize))]]`. So, there is explicit `le` for the size, and it stays constant across all pages. Am I using `fixed_size` in an unintended way?  I could just explicitly count the offset, and it would (I believe) make both the GUI and the testsuite happy, something like this:

```diff
diff --git a/patterns/keytab.hexpat b/patterns/keytab.hexpat
index 5bcdae1..39cc4d4 100644
--- a/patterns/keytab.hexpat
+++ b/patterns/keytab.hexpat
@@ -115,16 +115,16 @@ fn format_entry_contents(ref EntryContents e) {
 
 struct Entry {
     be s32 entry_size;
+    auto entry_start = $;
     if (entry_size == 0) {
         // end of entry list
-        return;
+        break;
     }
     else if (entry_size < 0) {
         u8 _gap[-entry_size];
         continue;
     }
     else {
-        EntryContents entry [[fixed_size(entry_size)]]; $ += entry_size;
+        EntryContents entry [[fixed_size(entry_size)]]; $ = entry_start + entry_size;
     }
 } [[format("format_entry")]];
 ```

Maybe I should ditch the `fixed_size` attribute completely then, as I'm not sure if it does anything in this case?

On a side note, whenever I try to set byte order globally, either with `#pragma` or with `std::core::set_endian`, the `std::time::to_local` and `std::time::to_utc` stop parsing timestamps and return Time structs filled with zeros. That's why I used explicit `be` mark on numbers.

WerWolv:
No, you're using fixed_size just fine. The difference probably comes from you running release ImHex and this uses the latest available commit. The $ += entry_size isn't necessary 

I will take a look at the implementation, it's possible you are running into a runtime bug

Mikle-Bond:
I tried the nightly build of ImHex, and yeah, `fixed_size` works just as I initially expected. Still, I've left the manual offset as is, on the "doesn't hurt" basis, as this way it works in 1.38.1 too. I also followed your advice and added keytab v1 format support, with `in` variable, and automatic guesses, so the tests should pass (at lest they passed on my machine this morning). Please, have another look.

WerWolv:
Thank you!

PR #578, opened 2026-09-04
patterns/hotline: Add Hotline client bookmark and resource file patterns

scottanderson:


WerWolv:
Please add the file to the readme and add test files if you can

PR #583, opened 2026-09-12
encodings: Generate .tbl files from Python stdlib codecs

scottanderson:
## Summary
- `scripts/generate_encodings.py` regenerates all of `encodings/` from Python's `codecs`/`encodings.aliases` instead of hand-maintained tables. `--check` (run in CI) keeps them in sync.
- Adds several missing encodings and fixes one that was mislabeled (`big5hkscs`), plus adds `cp949` (UHC) as its own entry, distinct from `EUC-KR` rather than mislabeled as it.
- Encodings that are exact supersets of another now `-include` it instead of repeating entries, found automatically. Near-matches (not exact) get their shared bytes factored into a table under `encodings/includes/`, only when the overlap is large enough (≥48 entries) to be worth a separate file. These de-dup tables are named by content hash, with friendly names for the recognizable ones.
- Every generated `.tbl` file now starts with a `# Generated by ...` comment.
- The README's encoding tables are generated too, and covered by the same `--check`.

## Testing
- [x] `python3 scripts/generate_encodings.py --check` passes
- [x] CI runs the same check


encodings/jis_x_0211.tbl
scottanderson, line 1: This file was just placeholder values, not really JIS_X_0211

scripts/generate_encodings.py
```diff
@@ -0,0 +1,592 @@
+#!/usr/bin/env python3
```
WerWolv, line 1: This should be called something like `generate_encodings.py`

scripts/generate_encodings.py
```diff
@@ -0,0 +1,592 @@
...
+ALGORITHMIC_ENCODINGS = [
+    ("UTF-8", "Unicode Transformation Format, 8-bit"),
+    ("UTF-16BE", "Unicode Transformation Format, 16-bit, big-endian"),
+    ("UTF-16LE", "Unicode Transformation Format, 16-bit, little-endian"),
+    ("UTF-32BE", "Unicode Transformation Format, 32-bit, big-endian"),
+    ("UTF-32LE", "Unicode Transformation Format, 32-bit, little-endian"),
+]
+
+# Hand-authored game-text tables; no codec covers them, but still listed below.
+HAND_AUTHORED_FILES = {
```
WerWolv, line 33: This is going to be annoying for anybody adding their own encodings... Maybe we should instead move all the auto generated ones into a sub folder so they can stay separate from the custom ones.
I'd mostly like to avoid people having to edit this file in order to submit their own encoding files

WerWolv (changes requested):

scripts/generate_encodings.py
scottanderson: My plan is to make the script ignore any tbl file that doesn't have a generated comment on the first line.

WerWolv: This table should be removed now. The only purpose it serves is generating the table in the readme which is easier to just keep updated there by hand instead of having to modify the script here

WerWolv (approved):

PR #584, opened 2026-09-12
feat: add osu replay pattern

Electrobytes:
Adds a pattern file for osu replay file format based from this [spec](https://osu.ppy.sh/wiki/en/Client/File_formats/osr_(file_format)). From what I researched, a replay file does not contain any media from the beatmap it rebroadcasts; therefore, the test file I've included should not have any copyrighted material.


WerWolv:
Thank you!

